PR #3, opened 2022-02-17
Fix protocol error when sending mqtt v4 msg to mqtt v5 client

JaylinYu:


JaylinYu:
just fixed the confliction, Please review, but it need to work with another in nanomq

src/sp/protocol/mqtt/mqtt_parser.c
```diff
@@ -1216,3 +1216,20 @@ nano_msg_ubsub_free(nano_pipe_db *db)
...
 	return;
 }
+
+/**
+ * @brief get property length from msg if any
+ * 
+ * @return uint32_t 
+ */
+uint32_t
+nni_mqtt_get_property_len(nni_msg *m)
```
wanghaEMQ, line 1226: This function looks like incomplete

wanghaEMQ (approved):

PR #13, opened 2022-02-26
* FIX [mqtt/transport&protocol] fix msg pointer & protocol error(wron…

JaylinYu:


JaylinYu (commented):
@wanghaEMQ have a look

PR #23, opened 2022-03-03
fix nanomq bridge 

JaylinYu:


alvin1221 (changes requested):
These changes of _mqtt_tcp.c_ and _brokert_tcp.c_ need to be synchronized to _mqtt_tls.c_ and _broker_tls.c_ .

alvin1221 (approved):

lee-emqx (approved):

src/supplemental/mqtt/mqtt_msg.c
```diff
@@ -674,4 +674,65 @@ nni_mqtt_msg_set_aio(nni_msg *msg, nni_aio *aio)
 {
 	nni_mqtt_proto_data *proto_data = nni_msg_get_proto_data(msg);
 	proto_data->aio = aio;
+}
+
+conn_param *
+nni_mqtt_msg_set_conn_param(nni_msg *msg)
```
wanghaEMQ, line 680: Such a ridiculous function name... I thought it was a get operation

JaylinYu: should be changed

PR #26, opened 2022-03-06
session expiry interval

### LINKED ISSUE #25: NanoMQ
broker does not return connack properly when recevie connect with session expiry interval

JaylinYu:
fix #25 & support session expiry

### DIFF
--- a/src/sp/protocol/mqtt/nmq_mqtt.c
+++ b/src/sp/protocol/mqtt/nmq_mqtt.c
@@ -193,14 +193,42 @@ nano_pipe_timer_cb(void *arg)
 	if (nng_aio_result(&p->aio_timer) != 0) {
 		return;
 	}
+	nni_mtx_lock(&p->lk);
+	// TODO pipe lock or sock lock?
 	if (npipe->cache) {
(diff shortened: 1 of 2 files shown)


lee-emqx (approved):

src/sp/protocol/mqtt/nmq_mqtt.c
```diff
@@ -193,14 +193,42 @@ nano_pipe_timer_cb(void *arg)
...
 	}
+	nni_mtx_lock(&p->lk);
+	// TODO pipe lock or sock lock?
 	if (npipe->cache) {
-		nni_sleep_aio(qos_duration * 1000, &p->aio_timer);
+		p->ka_refresh++;
 		// check session expiry interval
-		// npipe->cache = false;
-		// nni_pipe_close(p->pipe);
+		if (p->conn_param->session_expiry_interval > 0) {
```
wanghaEMQ, line 201: The case that session never expires seems not to be taken into account.
```
If the Session Expiry Interval is 0xFFFFFFFF (UINT_MAX), the Session does not expire.
```

PR #46, opened 2022-03-24
will msg expiry + will msg delay 

JaylinYu:
https://github.com/nanomq/nanomq/issues/313


JaylinYu (commented):
general msg expiry interval is on the way

alvin1221 (approved):

lee-emqx (approved):

PR #48, opened 2022-03-25
 * NEW [core] add nng_stream as arg of connect/disconnect callback.

JaylinYu:
we need nng_stream in other places for bridging


JaylinYu:
need more work to be compatible with NNG

PR #50, opened 2022-03-29
Replace nni_qos_db APIs compatible with `nni_id_hash_map` and `SQLite`

alvin1221:
implement #40


JaylinYu (approved):
need more test though

PR #85, opened 2022-04-16
FIX https://github.com/nanomq/NanoSDK/issues/39

JaylinYu:


JaylinYu:
sorry I have to pollute commit history today

PR #92, opened 2022-04-16
 * RM [core/aio] remove packetid from aio.

JaylinYu:
what ever happens today, stay today


JaylinYu:
stop polluting commit history

PR #103, opened 2022-04-26
Support client messages to be persisted.

alvin1221:


src/mqtt/protocol/mqtt/mqtt_client.c
```diff
@@ -483,10 +529,11 @@ mqtt_send_cb(void *arg)
 		mqtt_send_msg(c->saio, c);
 		return;
 	}
-	// Then those msg in nni_lmq
-	if (nni_lmq_get(&p->send_messages, &msg) == 0) {
+
```
JaylinYu, line 532: isn't a Macro control here better?

JaylinYu (changes requested):
plz also cache &p->sent_unack

JaylinYu (approved):

PR #104, opened 2022-04-28
* FIX [mqtt_parser] Fix the error in handling will msg.

wanghaEMQ:


JaylinYu (approved):
The rest part is mine

PR #115, opened 2022-05-06
 * FIX [] tmp fix for #595

JaylinYu:


src/sp/protocol/mqtt/mqtt_parser.c
```diff
@@ -354,7 +354,7 @@ utf8_check(const char *str, size_t len)
 			/* Not enough data */
 			return ERR_MALFORMED_UTF8;
 		}
-		for (j = 0; j < codelen - 1; j++) {
+		for (j = 0; j < codelen - 1 && len > 1; j++) {
```
lee-emqx, line 357: Should parentheses be added to increase code readability？

JaylinYu: sure 
I move this part to another PR, because this PR is reserve for wangha to fix the severe data racing on sub_ctx

PR #128, opened 2022-05-11
Fix issue https://github.com/emqx/nanomq/issues/584

lee-emqx:


include/nng/protocol/mqtt/mqtt.h
```diff
@@ -208,5 +208,6 @@ typedef enum {
 
 // Maximum Packet Size of broker
 #define NANO_MAX_RECV_PACKET_SIZE (2*1024*1024)  // TODO set in conf
+NNG_DECL void nni_msg_proto_set_property(nng_msg *msg, void *prop);
```
JaylinYu, line 211: let's simply use set_proto_data, discard this API. remember to replace that one in pub_handler

JaylinYu (changes requested):

include/nng/protocol/mqtt/mqtt.h
lee-emqx: Maybe it's hard to do, because set_proto_data need set function pointer(dup/free), these two functions is implemented at mqtt_codec.c.

JaylinYu (approved):

PR #133, opened 2022-05-20
Expose nng_atomic APIs.

alvin1221:


JaylinYu (approved):
I still prefer better not expose these

PR #142, opened 2022-05-31
send disconnect with error code

JaylinYu:


JaylinYu:
@alvin1221  plz return errcode when there are duplicated property in CONNECT

PR #153, opened 2022-06-17
fix memleak in qos_db_api

JaylinYu:
conn_param_clone is part of another PR in nanomq


alvin1221 (approved):
I'll check again later.

lee-emqx (approved):

PR #166, opened 2022-06-23
 * MDF [mqtt_client] rename bridge_mode in conf struct to enable (mul…

JaylinYu:
just a rename , but necessary for https://github.com/emqx/nanomq/pull/680


alvin1221 (approved):
make sense

lee-emqx (approved):

PR #173, opened 2022-06-28
 * MDF [nmq_mqtt] replace mutex lock with rwlock

JaylinYu:


JaylinYu:
but i am not sure it is the right move

PR #182, opened 2022-07-05
Move handling of subinfol from nmq_mqtt to transport layer

wanghaEMQ:


JaylinYu (commented):
emulate send_start_v5

JaylinYu (approved):
plz make subtopic safe, be cautious with strchr, will move your pointer nowhere!

PR #190, opened 2022-07-08
new cb for nanosdk

JaylinYu:
will do the same for nanosdk repo


JaylinYu:
gonna do the same for nanosdk

PR #236, opened 2022-08-09
Add repub conf parse function.

lee-emqx:


src/sp/transport/mqtt/broker_tcp.c
```diff
@@ -441,6 +441,11 @@ nmq_tcptran_pipe_qos_send_cb(void *arg)
 	msg  = nni_aio_get_msg(qsaio);
 	type = nni_msg_cmd_type(msg);
 
+	if (p->closed) {
+		nni_msg_free(msg);
+		goto exit;
+
```
JaylinYu, line 447: better delete this line and move checker upward before the lock

JaylinYu (approved):

src/sp/transport/mqtt/broker_tcp.c
JaylinYu: I will make it better later

PR #255, opened 2022-08-20
MQTT over QUIC bridging ver 1

JaylinYu:


JaylinYu:
better to create a buffer to store inflight data return by msquic

JaylinYu:
no longer valid

PR #267, opened 2022-09-02
online/offline msg of QUIC bridge

JaylinYu:


src/supplemental/quic/quic_api.c
```diff
@@ -885,7 +877,7 @@ quic_strm_recv_cb(void *arg)
 				2 * nni_lmq_cap(&qstrm->recv_messages))) {
 				// memory error
 				nni_msg_free(qstrm->rxmsg);
-				printf("msg dropped due to no more memory!\n");
+				nni_println("msg dropped due to no more memory!\n");
```
lee-emqx, line 880: Why do not use alvin's log?

JaylinYu: fine

PR #280, opened 2022-09-09
Move MQTT ping to quic layer from mqtt layer and add reconnect mechanism

wanghaEMQ:
1. Move MQTT ping to quic layer from mqtt layer
2. Add a quic reconnect mechanism which relies on MQTT ping


JaylinYu:
cannot get keepalive interval from conf/connmsg.

JaylinYu (approved):
just fixed a minor bug of yours

PR #307, opened 2022-09-30
Fix wrong ref number of quic sock

JaylinYu:


JaylinYu (commented):
event msg still not valid in hybrid bridging, requires a clean reopen of sock and ctxs

wanghaEMQ (commented):
good job!

PR #321, opened 2022-10-24
 * FIX [CMakeLIst] only compile mqtt_quic.c when QUIC is in enabled c…

JaylinYu:


include/nng/mqtt/mqtt_quic.h
```diff
@@ -17,7 +17,7 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
-
+#if defined(SUPP_QUIC)
```
wanghaEMQ, line 20: Actually, this file would not be included if NNG_PROTO_MQTT_QUIC_CLIENT option is not ON

wanghaEMQ (approved):

include/nng/mqtt/mqtt_quic.h
JaylinYu: isnt this what we want?

PR #322, opened 2022-10-25
Fix  set conn_param = NULL unexpected in mqtt_tls.c

alvin1221:


lee-emqx (approved):

JaylinYu (approved):
thx. is this mean websocket over TLS is also working fine?

PR #375, opened 2022-12-10
Support max_send_queue_length param.

lee-emqx:


JaylinYu (approved):
seems good , but why dont you  make it valid in nanomq by setopt?

PR #408, opened 2023-01-11
Add fnv1a & crc & crc32 & crc32c hash algorithms supported.

wanghaEMQ:
Follow hash algorithms supported. And support config with polynomials.
* fnv1a
* crc
* crc32
* crc32c


JaylinYu (approved):
nice

PR #431, opened 2023-02-11
* FIX [mqtt_codec] fix https://github.com/emqx/nanomq/issues/1042

JaylinYu:


JaylinYu:
validate property length before decoding it!

alvin1221 (approved):

PR #444, opened 2023-02-20
Add utils.c to unify `fatal()` function

alvin1221:


lee-emqx (approved):

JaylinYu:
it is not finished yet. All modules like bridge/web_server/web_hook are still using internal func.
or you wanna do this in another pr?

JaylinYu (approved):

PR #448, opened 2023-02-23
Change the way of openssl dependency.

lee-emqx:


lee-emqx:
Use Ninja option can be compile parallel, but 'make -j n' does't work now. Because dependencies can not work in order, but Ninja can keep dependencies in order.

lee-emqx (commented):
Use Ninja option can be compile parallel, but 'make -j n' does't work now. Because dependencies can not work in order, but Ninja can keep dependencies in order.

PR #469, opened 2023-03-22
Fix a dead lock in dbtree_delete_retain in mqtt_db.c

moonZHH:
To prevent the dead lock in dbtree_delete_retain, call nni_rwlock_unlock after mem_free instead of before mem_free (referring to the correct implementation in dbtree_delete_client).

fixes #<issue number> <issue synopsis>

<Comments describing your change. Not all changes need this.>

Note that the above format should be used in your git commit comments.
You agree that by submitting a PR, you have read and agreed to our
contributing guidelines.


lee-emqx (approved):

lee-emqx:
@moonZHH Thanks for your fix.

PR #543, opened 2023-04-28
 Fix memleak in quic

Hermann0222:


Hermann0222:
msquic is too heavy for ci.

JaylinYu (approved):

PR #564, opened 2023-05-17
* FIX [nmq_mqtt] pipe reaper needs a bigger lock #1180 of NanoMQ

JaylinYu:


JaylinYu:
This solves https://github.com/emqx/nanomq/issues/1180, but might dragging down the performance

PR #567, opened 2023-05-19
* FIX [mqtt/websocket] a potential blocking error https://github.com/…

JaylinYu:


JaylinYu:
related to https://github.com/emqx/nanomq/issues/1210

PR #570, opened 2023-05-25
Sync NanoSDK to NanoNNG.

lee-emqx:


JaylinYu (approved):
How about a flag not a new protocol layer file

JaylinYu (approved):
How about a flag not a new protocol layer file

PR #577, opened 2023-06-06
Modify system part config changes.

lee-emqx:


lee-emqx:
> hide this para not delete

OK

JaylinYu (approved):

PR #594, opened 2023-06-29
* MDF [mqtt_msg] init objs of proto_Data struct

JaylinYu:
sync fix from nanosdk

### DIFF
--- a/src/supplemental/mqtt/mqtt_msg.c
+++ b/src/supplemental/mqtt/mqtt_msg.c
@@ -16,6 +16,18 @@ nni_proto_data_init(nni_mqtt_proto_data *proto_data)
 {
 	proto_data->var_header.connect.conn_flags.clean_session = true;
 	proto_data->var_header.connect.keep_alive = 30;
+	proto_data->var_header.connack.properties = NULL;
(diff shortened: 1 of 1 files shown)


src/supplemental/mqtt/mqtt_msg.c
```diff
@@ -16,6 +16,18 @@ nni_proto_data_init(nni_mqtt_proto_data *proto_data)
 {
 	proto_data->var_header.connect.conn_flags.clean_session = true;
 	proto_data->var_header.connect.keep_alive = 30;
+	proto_data->var_header.connack.properties = NULL;
```
lee-emqx, line 19: It's a union here, the initialize seems improper?

JaylinYu: still a pointer here

lee-emqx: I guess we don't need this function, all we need is just `memset` proto_data is zero.

JaylinYu: you are right

JaylinYu:
TODO: memset proto_data to /0.

PR #599, opened 2023-07-03
Which fixed https://github.com/emqx/nanomq/issues/1329

wanghaEMQ:
* Fix the double free in qos msg for quic.


JaylinYu:
a better solution in #600

Issue #573, opened 2023-06-03
Massive quic qos 2 message pub will cause heap buffer flow.


lee-emqx:
```
=================================================================
==980499==ERROR: AddressSanitizer: attempting double-free on 0x603000092ef0 in thread T37:
    #0 0x7f9d8e2b4517 in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:127
    #1 0x5651a63b6806 in nni_free /home/lee/workspace/nanomq/nng/src/platform/posix/posix_alloc.c:33
    #2 0x5651a6409127 in mqtt_msg_content_free /home/lee/workspace/nanomq/nng/src/supplemental/mqtt/mqtt_codec.c:370
    #3 0x5651a640943e in nni_mqtt_msg_free /home/lee/workspace/nanomq/nng/src/supplemental/mqtt/mqtt_codec.c:425
    #4 0x5651a63a0468 in nni_msg_free /home/lee/workspace/nanomq/nng/src/core/message.c:462
    #5 0x5651a63c8df5 in mqtt_quic_recv_cb /home/lee/workspace/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic.c:1139
    #6 0x5651a63b13cf in nni_task_exec /home/lee/workspace/nanomq/nng/src/core/taskq.c:144
    #7 0x5651a6393d24 in nni_aio_finish_impl /home/lee/workspace/nanomq/nng/src/core/aio.c:454
    #8 0x5651a6393da6 in nni_aio_finish_sync /home/lee/workspace/nanomq/nng/src/core/aio.c:469
    #9 0x5651a6438a20 in quic_pipe_recv_cb /home/lee/workspace/nanomq/nng/src/supplemental/quic/quic_api.c:1207
    #10 0x5651a63b13cf in nni_task_exec /home/lee/workspace/nanomq/nng/src/core/taskq.c:144
    #11 0x5651a6393d24 in nni_aio_finish_impl /home/lee/workspace/nanomq/nng/src/core/aio.c:454
    #12 0x5651a6393da6 in nni_aio_finish_sync /home/lee/workspace/nanomq/nng/src/core/aio.c:469
    #13 0x5651a6433d70 in quic_strm_cb /home/lee/workspace/nanomq/nng/src/supplemental/quic/quic_api.c:468
[...]

PR #604, opened 2023-07-04
Fix a hidden UAF bug in quic transport

wanghaEMQ:
Which fixed https://github.com/nanomq/NanoNNG/issues/573
* Add rraio finish in a aync way.
* Make rraio in mtx guard.
* Sequence the order to run rraio.


JaylinYu (approved):
Bravo!

lee-emqx (approved):

PR #611, opened 2023-07-10
* FIX [quic_api] FIx the error in quic_api.

wanghaEMQ:


src/supplemental/quic/quic_api.c
```diff
@@ -1507,6 +1508,7 @@ quic_close()
 			MsQuic->RegistrationClose(registration);
 		}
 		MsQuicClose(MsQuic);
+		MsQuic = NULL;
```
JaylinYu: better prederve this

JaylinYu (approved):

PR #614, opened 2023-07-11
Fix the null mtx in qstrm.

wanghaEMQ:


JaylinYu (approved):
......................

PR #620, opened 2023-07-17
Fix the double free in connmsg when ep finite.

wanghaEMQ:
* FIX [mqtt_tcp] Fix the double free in connmsg when ep finite.

### DIFF
--- a/src/mqtt/transport/tcp/mqtt_tcp.c
+++ b/src/mqtt/transport/tcp/mqtt_tcp.c
@@ -1137,13 +1137,15 @@ mqtt_tcptran_ep_fini(void *arg)
 
 	nni_mtx_lock(&ep->mtx);
 	ep->fini = true;
-	if (ep->connmsg)
-		nni_msg_free(ep->connmsg);
 	if (ep->refcnt != 0) {
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


src/mqtt/transport/tcp/mqtt_tcp.c
```diff
@@ -1137,13 +1137,15 @@ mqtt_tcptran_ep_fini(void *arg)
...
-		nni_msg_free(ep->connmsg);
 	if (ep->refcnt != 0) {
 		nni_mtx_unlock(&ep->mtx);
 		return;
 	}
 	nni_mtx_unlock(&ep->mtx);
+	// Free connmsg once
+	if (ep->connmsg)
+		nni_msg_free(ep->connmsg);
+
```
RanMaoyi, line 1148: Maybe do this in `ep->mtx` is better?

RanMaoyi (approved):
LGTM

src/mqtt/transport/tcp/mqtt_tcp.c
wanghaEMQ: I think the refcnt has ensured that the free only executes once.

lee-emqx (approved):

JaylinYu (approved):
does it related to previous bridging reload bug?

PR #627, opened 2023-07-20
Enable server verification by default for mqtt_async tls demo.

lee-emqx:


JaylinYu (approved):
not sure what happened here

RanMaoyi (approved):

lee-emqx:
> not sure what happened here

I just want it to be correct.

PR #645, opened 2023-08-03
Fix some errors in transport layer and add transport layer code to cmake

wanghaEMQ:
* Fix some building errors in transport layer.
* Add transport layer code to cmake


JaylinYu (approved):
Nicely done

PR #659, opened 2023-08-10
* FIX [rule] fix memleak of rule_mutex.

Hermann0222:


RanMaoyi (commented):
Why introduce rule_mutex, which doesn't seem to lock/unlock, just alloc and free

Hermann0222:
> Why introduce rule_mutex, which doesn't seem to lock/unlock, just alloc and free

It will be used in nanomq when rule engine is on. The mutex used in nanomq rule engine now will cause memleak and data race. this pr will sync to nanomq soon.

RanMaoyi (approved):

lee-emqx (approved):

PR #676, opened 2023-08-23
* FIX [ws] Fix a timing issue

RanMaoyi:
If recv_cb executes before accept_cb, ep_aio will be empty, and the uaio in recv_cb will not be assigned correctly, which will cause the listener's accaio to fail to finish properly, resulting in a ws link establishment failure problem.

fixes (https://github.com/emqx/nanomq/issues/1210)

image


JaylinYu:
@Hermann0222 lets make a coverage test on ws

JaylinYu (approved):
ok cool;
my bad

Hermann0222:
> @Hermann0222 lets make a coverage test on ws

Roger

PR #678, opened 2023-08-24
* FIX [mqtt_parser] Error return is not performed when the username o…

RanMaoyi:
EMQX websocket client set 0x80 and 0x40 in conn_flag, but no username or password is seted. Which will cause emqx show connected then disconnected soon.

fixes https://github.com/emqx/nanomq/issues/1420


JaylinYu:
this is not aligned with MQTT standard. this is their bug not ours

PR #684, opened 2023-08-30
* FIX [conf test] a patch to fix CI fail in windows.

OdyWayne:


cmake/NNGHelpers.cmake
```diff
@@ -111,7 +111,7 @@ function(nng_test NAME)
                 ${PROJECT_SOURCE_DIR}/src
                 ${PROJECT_SOURCE_DIR}/include)
         add_test(NAME ${NNG_TEST_PREFIX}.${NAME} COMMAND ${NAME} -t -v)
-        set_tests_properties(${NNG_TEST_PREFIX}.${NAME} PROPERTIES TIMEOUT 180)
+        set_tests_properties(${NNG_TEST_PREFIX}.${NAME} PROPERTIES TIMEOUT 60)
```
wanghaEMQ, line 114: Not recommended... This will increase the cost of rebasing upstream

wanghaEMQ (commented):
Cool

wanghaEMQ (approved):

JaylinYu (approved):
DK what happened?

wanghaEMQ:
> DK what happened?

On windows. CI spent too much time on reading the configuration file. And always failed at the end.

OdyWayne:
conf parse has been using a posix api which windows do not support.

RanMaoyi:
> conf parse has been using a posix api which windows do not support.

which api?

OdyWayne:
> > conf parse has been using a posix api which windows do not support.
> 
> which api?

I was wrong, the real problem is that the format of the file path is different in windows.

PR #695, opened 2023-09-18
a patch to fix CI failure in windows

OdyWayne:


RanMaoyi (approved):

OdyWayne:
> A little advice, if you can avoid Shotgun Surgery is better

good to know. But this patch is just turned out not the case, so I decided to just revert it with minor changes.

wanghaEMQ (approved):

lee-emqx (approved):

PR #701, opened 2023-09-22
Fix a crash when localtopic/remotetopic is not set

wanghaEMQ:
Which fixes https://github.com/emqx/nanomq/issues/1457


OdyWayne (approved):

lee-emqx (approved):

RanMaoyi (approved):

JaylinYu (approved):
Doesn't occur on old config?

OdyWayne:
> Doesn't occur on old config?

works fine with old configuration.

PR #703, opened 2023-09-26
conf boost

OdyWayne:
this is also a test for CI.


RanMaoyi (approved):

src/supplemental/nanolib/conf.c
```diff
@@ -2388,7 +2392,12 @@ conf_bridge_node_parse_subs(
 			}
 		}
 	}
-
+	if (get_remote_topic) {
```
OdyWayne, line 2395: I think it just fix the potential memleak mentioned in line 2297 and line 2308, right?

wanghaEMQ (commented):
lgtm

lee-emqx (approved):

PR #724, opened 2023-11-03
NOTE

wanghaEMQ:
A new issue.
Nanomq will create multiple stream which number is more than it should create.


wanghaEMQ:
A efficient solution has been provided at https://github.com/nanomq/NanoNNG/pull/732

Issue #726, opened 2023-11-07
Quic bridge send aio isn't be finished when multi stream is enabled


wanghaEMQ:
**Describe the bug**
```
2023-11-07 05:36:03 [85819] INFO  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:848  msquic_connection_cb: QUIC_CONNECTION_EVENT_DATAGRAM_STATE_CHANGED
2023-11-07 05:36:03 [85819] INFO  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:755  msquic_connection_cb: [conn][0x620000001080] is Connected. Resumed Session 0
2023-11-07 05:36:03 [85819] INFO  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:851  msquic_connection_cb: QUIC_CONNECTION_EVENT_STREAMS_AVAILABLE
2023-11-07 05:36:03 [85819] INFO  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:851  msquic_connection_cb: QUIC_CONNECTION_EVENT_STREAMS_AVAILABLE
2023-11-07 05:36:03 [85819] INFO  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:996  msquic_strm_cb: QUIC_STREAM_EVENT_START_COMPLETE [0x618000040080] ID: 0 Status: 0
2023-11-07 05:36:03 [85790] INFO  /home/wangha/Documents/nanomq/nng/src/mqtt/transport/quic/mqtt_quic.c:1462  mqtt_quictran_dial_cb: main pipe 0x610000000240
2023-11-07 05:36:03 [85819] WARN  /home/wangha/Documents/nanomq/nng/src/supplemental/quic/msquic_dial.c:815  msquic_connection_cb: [conn][0x620000001080] Resumption ticket received (1764 bytes):

2023-11-07 05:36:03 [85796] INFO  /home/wangha/Documents/nanomq/nanomq/bridge.c:739  bridge_quic_connect_cb: Bridge client connected! RC [0]
2023-11-07 05:36:03 [85796] INFO  /home/wangha/Documents/nanomq/nanomq/bridge.c:740  bridge_quic_connect_cb: Local ip4 address [0.0.0.0] port [0]
[...]

PR #727, opened 2023-11-07
Fix the error that aio is not finished.

wanghaEMQ:
Fixed <https://github.com/nanomq/NanoNNG/issues/726>
But still some aios are not been finished.


RanMaoyi (approved):

OdyWayne (approved):

src/mqtt/transport/quic/mqtt_quic.c
```diff
@@ -1601,6 +1598,9 @@ mqtt_quictran_ep_connect(void *arg, nni_aio *aio)
 	mqtt_quictran_ep *ep = arg;
 	int              rv;
 
+	// XXX Reset the d_started flag to allow eatablish multistream fast
+	nni_atomic_flag_reset(&ep->ndialer->d_started);
```
JaylinYu, line 1602: why do u put it here?

wanghaEMQ: It was put to the callback of connection. But this may lead to nni_dialer_start returning an error if nni_dialer_start is too fast. That's why it is put here.

PR #730, opened 2023-11-10
Export QUIC stream priority api

RanMaoyi:
You can set dialer quic priority before dialer_start to create a stream with the specified priority.
```
nni_dialer_setopt(npipe->p_dialer, NNG_OPT_MQTT_QUIC_PIRORITY, &val, sizeof(int), NNI_TYPE_INT32);
```


src/mqtt/protocol/mqtt/mqtt_quic_client.c
```diff
@@ -1749,6 +1749,12 @@ mqtt_quic_ctx_send(void *arg, nni_aio *aio)
 			if (pub_pipe == NULL) {
 				log_info("create new pub stream");
 				nni_lmq_put(s->topic_lmq, msg);
+				/*
+				 * You can set dialer quic priority before
+				 * dialer_start to create a stream with the
+				 * specified priority.
+				 */
+				//nni_dialer_setopt(npipe->p_dialer, NNG_OPT_MQTT_QUIC_PIRORITY, &val, sizeof(int), NNI_TYPE_INT32);
```
wanghaEMQ, line 1757: Is the priority property just for publish stream?

RanMaoyi: For now, only publish streams.

But it can get complicated later, so this is just exposing the api. At the same time, I did some tests, but the test effect is not very significant.

include/nng/nng.h
```diff
@@ -737,6 +737,8 @@ NNG_DECL nng_listener nng_pipe_listener(nng_pipe);
 #define NNG_OPT_QUIC_SEND_IDLE_TIMEOUT "quic-send-idle-timeout"
 #define NNG_OPT_QUIC_INITIAL_RTT_MS "quic-initial-ms"
 #define NNG_OPT_QUIC_MAX_ACK_DELAY_MS "quic-max-ack-delay-ms"
+#define NNG_OPT_MQTT_QUIC_PIRORITY "quic-mqtt-stream-priority"
+#define NNG_OPT_QUIC_PRIORITY "quic-stream-priority"
```
wanghaEMQ, line 741: `NNG_OPT_MQTT_QUIC_PIRORITY` is unnecessary. Transport layer don't need to know it's a priority stream. You could set priority property directly to quic stream layer.

wanghaEMQ (approved):

src/mqtt/protocol/mqtt/mqtt_quic_client.c
wanghaEMQ: In my opinion. Subscribe request should have a priority that is not too low. (It could be a fixed priority value)

include/nng/nng.h
RanMaoyi: I know it's easy to set priority property directly to Quic stream layer, but if you try to access the transport dialer from the protocol layer, it is not reasonable and will cause a compilation error like this.
```
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c: In function ‘mqtt_quic_ctx_send’:
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c:1758:33: error: unknown type name ‘mqtt_quictran_ep’
 1758 |                                 mqtt_quictran_ep *ep = (mqtt_quictran_ep *)npipe->p_dialer;
      |                                 ^~~~~~~~~~~~~~~~
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c:1758:57: error: ‘mqtt_quictran_ep’ undeclared (first use in this function)
 1758 |                                 mqtt_quictran_ep *ep = (mqtt_quictran_ep *)npipe->p_dialer;
      |                                                         ^~~~~~~~~~~~~~~~
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c:1758:57: note: each undeclared identifier is reported only once for each function it appears in
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c:1758:75: error: expected expression before ‘)’ token
 1758 |                                 mqtt_quictran_ep *ep = (mqtt_quictran_ep *)npipe->p_dialer;
      |                                                                           ^
/home/maoyi/github/emqx/nanomq/nng/src/mqtt/protocol/mqtt/mqtt_quic_client.c:1759:57: error: request for member ‘dialer’ in something not a structure or union
 1759 |                                 nni_stream_dialer_set(ep->dialer, NNG_OPT_QUIC_PRIORITY, &tmp, sizeof(int), NNI_TYPE_INT32);
      |                                                         ^~

```

lee-emqx (approved):

include/nng/nng.h
wanghaEMQ: Change tack and create a nng_dialer and set the id of it then call `nng_dialer_set_xxx`.
Just like `nanomq_cli/client.c:1549 nng_dialer_set_bool(dialer, NNG_OPT_QUIC_ENABLE_0RTT, true);` did.

RanMaoyi: Protocol layer code is just an example of how to set quic stream priority, so if create a new dialer before set priority Quic stream is unnecessary.(The change is not consistent with quic's multi-stream design....)

And in this PR we exposed two api to users who want to set priority of Quic stream, how to use and which api to use is upon users.

In future work, we will consider quic multi-stream priority scheduling, not only based on topic priority.

wanghaEMQ: of course yes. you are right.

JaylinYu (approved):
modify the same int priority each time when create a new stream?

PR #733, opened 2023-11-16
Introduce ring buffer!

RanMaoyi:


wanghaEMQ (approved):

wanghaEMQ:
Cooooool

lee-emqx (approved):

OdyWayne (approved):

include/nng/supplemental/nanolib/ringbuffer.h
```diff
@@ -0,0 +1,35 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+struct ringBufferMsg {
+	void *data;
+	/* TTL of each message */
+	unsigned long long expiredAt;
```
JaylinYu, line 7: whats the plan to make this valid

JaylinYu (approved):
not even relavant with NNG

JaylinYu:
CMakelist is required

include/nng/supplemental/nanolib/ringbuffer.h
RanMaoyi: - When the program is initialized, a timer is created to notify when a message expires.

- When a new message arrives, the expiration time of the message is set based on the ttl, and it is necessary to check whether the message expiration time is less than the expiration time of the timer. If it is less than, the timer is turned off and a new timer is created for the expiration time of the message.

- When the message is consumed normally, the timer is checked to see if it is set for the message, and if so, the timer is turned off, and the next message expiration time is found and a new timer expiration time is set.

- When the timer fires, the message is processed, and when the processing is complete, the next message expiration time is found, and a new timer is created.

However, the problem is that our data structure is an array instead of a linked list, and message dropping can result in empty data, thus wasting space. I am implementing a new ringbuffer data structure based on linked lists.

RanMaoyi:
> CMakelist is required

Updated

JaylinYu:
Better add corresponding unit test

PR #743, opened 2023-11-21
Support read backoff max from configuration.

wanghaEMQ:


include/nng/supplemental/nanolib/conf.h
```diff
@@ -248,6 +248,7 @@ struct conf_bridge_node {
 	uint8_t      proto_ver;
 	uint16_t     port;
 	uint16_t     keepalive;
+	uint16_t     backoff_max;
```
RanMaoyi, line 251: Could uint16_t be too small and nni_duration be a better choice?

RanMaoyi (approved):

OdyWayne (approved):

include/nng/supplemental/nanolib/conf.h
JaylinYu: a solid concern. but nobody wants a loooooooog sleep backoff

JaylinYu (approved):
a little bit redundant

wanghaEMQ:
> a little bit redundant

Where?

PR #752, opened 2023-12-05
[WIP] Ringbuffer with key and export some api for get messages from exchange_client

RanMaoyi:
Work in process
Need more tests


JaylinYu (approved):
thats all?

wanghaEMQ (approved):

RanMaoyi:
> thats all?

@JaylinYu Currently, there are still some header file dependency issues that need to be resolved. The main issue is that 'conf.h' includes 'exchange_client.h,' and 'exchange_client.h' has a dependency on 'ringbuffer.h,' which includes 'nng_impl.h.' On the NanoMQ side, including 'nng_impl.h' causes header file path errors. Additionally, there are some test cases that still require confirmation of lock order.

PR #755, opened 2023-12-07
tlstran_pipe_stop should free cparam

lkyjason:
fixes #1519 Several memleaks in TLS

NanoMQ issue: https://github.com/nanomq/nanomq/issues/1519

p->tcp_cparam that is allocated in tlstran_pipe_nego_cb is not freed by tlstran_pipe_stop, instead the variable is directly set to NULL resulting in memory leak.

### DIFF
--- a/src/sp/transport/mqtts/broker_tls.c
+++ b/src/sp/transport/mqtts/broker_tls.c
@@ -156,10 +156,6 @@ tlstran_pipe_stop(void *arg)
 {
 	tlstran_pipe *p = arg;
 	log_trace(" ###### tlstran_pipe_stop ###### ");
-	if (p->tcp_cparam) {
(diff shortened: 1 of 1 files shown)


OdyWayne:
Thank you for your contribution! But we usually do frees in pipe_fini not pipe_stop. We will appreciate it if you can move all these to tlstran_pipe_fini, just like what we do in tcptran_pipe_fini. Or should I do it for you?

OdyWayne (changes requested):
better do this in pipe_fini

JaylinYu (approved):

JaylinYu:
Some modification we did on broker_tcp needs to be synced to broker_tls

OdyWayne:
well done

lkyjason:
Thank you @OdyWayne

PR #761, opened 2023-12-11
One exchange_client has one exchange and one exchange own one ringbuffer && void *** param for get msgs return

RanMaoyi:
The api for get msgs is still need to discuss. A new api or use nng_send_aio?

The broken UT need to fix, wip....

NEW config seems like
image

And an exchange demo can see https://github.com/nanomq/nanomq/tree/moi_exchange_demo


wanghaEMQ (approved):

JaylinYu (approved):

JaylinYu:
ok remember to fix UT

PR #777, opened 2023-12-21
* NEW [conf_ver2] disable tcp by not config it

OdyWayne:


JaylinYu (approved):
well done

wanghaEMQ (approved):

src/supplemental/nanolib/conf_ver2.c
```diff
@@ -351,6 +351,8 @@ conf_basic_parse_ver2(conf *config, cJSON *jso)
 			hocon_read_address_base(config, url, "bind", "nmq-tcp://", jso_tcp);
 			config->enable = true;
 			hocon_read_bool_base(config, enable, "enable", jso_tcp);
+		} else {
+			config->enable = false;
```
lee-emqx, line 355: if default value of `config->enable`  is false?

lee-emqx (approved):

RanMaoyi (approved):

src/supplemental/nanolib/conf_ver2.c
OdyWayne: No, if there is no configuration for `listeners.tcp`. Then `config->enable` will be false.

PR #786, opened 2023-12-22
Support  conf encryption option.

lee-emqx:


JaylinYu (approved):

wanghaEMQ (commented):
Coool

wanghaEMQ:
How about giving an example about this option in example.conf? @lee-emqx

PR #794, opened 2024-01-03
enable override retain flag

OdyWayne:


include/nng/supplemental/nanolib/conf.h
```diff
@@ -178,13 +178,16 @@ struct conf_websocket {
 
 typedef struct conf_websocket conf_websocket;
 
+#define NORETAIN 2 // default retain flag value, none 0, 1
```
wanghaEMQ, line 181: Need a prefix

wanghaEMQ (approved):

PR #811, opened 2024-01-16
Support read plugin from config file

RanMaoyi:
For example:

```
plugin {
    libs = [{
        path = "/path/to/plugin_alpha.so"
    }, {
        path = "/path/to/xxx.so"
    }]
}
```

### DIFF
--- a/include/nng/supplemental/nanolib/conf.h
+++ b/include/nng/supplemental/nanolib/conf.h
@@ -310,6 +310,12 @@ struct conf_exchange {
 	conf_exchange_node **nodes;
 };
 
+typedef struct conf_plugin conf_plugin;
+struct conf_plugin {
+	char **path;
+	size_t path_sz;
+};
+
 typedef enum {
 	UNCOMPRESSED,
 	SNAPPY,
@@ -528,6 +534,9 @@ struct conf {
 	conf_bridge          aws_bridge;	// AWS IoT Core
 	conf_exchange        exchange;
 	conf_parquet         parquet;
+#if defined(SUPP_PLUGIN)
+	conf_plugin          plugin;
+#endif
 	conf_web_hook        web_hook;
 #if defined(ENABLE_LOG)
 	conf_log  log;
--- a/include/nng/supplemental/nanolib/conf.h
+++ b/include/nng/supplemental/nanolib/conf.h
@@ -311,9 +311,14 @@ struct conf_exchange {
 };
 
 #if defined(SUPP_PLUGIN)
+typedef struct conf_plugin_lib conf_plugin_lib;
+struct conf_plugin_lib {
+	char *path;
+};
+
 typedef struct conf_plugin conf_plugin;
 struct conf_plugin {
-	char **path;
+	struct conf_plugin_lib **libs;
 	size_t path_sz;
 };
 #endif
--- a/include/nng/supplemental/nanolib/conf.h
+++ b/include/nng/supplemental/nanolib/conf.h
@@ -310,11 +310,13 @@ struct conf_exchange {
 	conf_exchange_node **nodes;
 };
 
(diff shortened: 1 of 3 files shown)


OdyWayne (approved):

src/supplemental/nanolib/conf_ver2.c
```diff
@@ -1028,10 +1028,45 @@ conf_bridge_parse_ver2(conf *config, cJSON *jso)
 	return;
 }
 
+#if defined(SUPP_PLUGIN)
+static void
+conf_plugin_parse_ver2(conf *config, cJSON *jso)
+{
+	cJSON *plugin = hocon_get_obj("plugin", jso);
```
lee-emqx, line 1035: Maybe use  hocon_get_obj in this way `cJSON *libs = hocon_get_obj("plugin.libs", jso)` is better.

include/nng/supplemental/nanolib/conf.h
```diff
@@ -310,6 +310,14 @@ struct conf_exchange {
 	conf_exchange_node **nodes;
 };
 
+#if defined(SUPP_PLUGIN)
+typedef struct conf_plugin conf_plugin;
+struct conf_plugin {
+	char **path;
+	size_t path_sz;
+};
+#endif
```
wanghaEMQ, line 319: Structure `conf_plugin` should align with the structure of it in config.

for example. the structure of plugin in configuration is 
```
plugin {
    libs = [{
        path = "/path/to/plugin_alpha.so"
    }, {
        path = "/path/to/xxx.so"
    }]
}
```

Then the structure of conf_plugin should be 
```
struct conf_plugin_lib {
    char *path;
}
struct conf_plugin {
    struct conf_plugin_lib **libs;
    size_t len;
}
```

src/supplemental/nanolib/conf_ver2.c
```diff
@@ -1028,10 +1028,45 @@ conf_bridge_parse_ver2(conf *config, cJSON *jso)
...
+	config->plugin.path_sz  = 0;
+
+	libs = hocon_get_obj("libs", plugin);
+
+	cJSON_ArrayForEach(lib, libs) {
+		cJSON *jso_path = cJSON_GetObjectItem(lib, "path");
+		if (jso_path) {
+			if (cJSON_IsString(jso_path)) {
+				if (jso_path->valuestring != NULL) {
+					log_error("jso_path->valuestring: %s", jso_path->valuestring);
```
wanghaEMQ, line 1049: Donot forget to remove the log when PR is already. Maybe changing its log level to info is also fine

wanghaEMQ (changes requested):

src/supplemental/nanolib/conf_ver2.c
RanMaoyi: TKS, removed

include/nng/supplemental/nanolib/conf.h
RanMaoyi: Got it, done

src/supplemental/nanolib/conf_ver2.c
RanMaoyi: Good suggestion, already done

wanghaEMQ (approved):

wanghaEMQ:
Well done

wanghaEMQ:
How about adding a print_plugin_conf function at your next PR? @RanMaoyi
So when plugin is configured then the content will be printed at `conf.c:1345`.

RanMaoyi:
> How about adding a print_plugin_conf function at your next PR? @RanMaoyi So when plugin is configured then the content will be printed at `conf.c:1345`.

ok, done

JaylinYu (commented):
how about the old conf

PR #842, opened 2024-01-30
Add vin related interfaces. 

lee-emqx:


wanghaEMQ (approved):

JaylinYu (commented):
what if there is a null value?

PR #881, opened 2024-02-29
* FIX [parquet] file_path UAF error

RanMaoyi:


xinyi-xs (approved):

wanghaEMQ (approved):
Will it cause memory leakage?

RanMaoyi:
> Will it cause memory leakage?

file_path will be freed in remove_old_file()

PR #883, opened 2024-03-01
exchange_server supports dynamically modifying the fullOp of the ringbus in sock_recv

RanMaoyi:


xinyi-xs (approved):

src/supplemental/nanolib/ringbuffer/ringbuffer.c
```diff
@@ -695,6 +695,18 @@ static int put_msgs_to_aio(ringBuffer_t *rb, nng_aio *aio)
 	return 0;
 }
 
+int ringBuffer_fullOp(ringBuffer_t *rb, enum fullOption fullOp)
```
wanghaEMQ, line 698: How about renaming to ringBuffer_set_fullOp

wanghaEMQ (approved):

src/supplemental/nanolib/ringbuffer/ringbuffer.c
RanMaoyi: you are right, rename to ringBuffer_set_fullOp

PR #894, opened 2024-03-14
Fixed parquet remove file memory leak.

xinyi-xs:


JaylinYu:
also apply to master branch

PR #895, opened 2024-03-19
Handling Ping & QoS12 msgs in protocol layer is supported

wanghaEMQ:
* Support handle ping msg in protocol layer
* Transport layer will ignore the qos msgs and let protocol layer to handle them.
* Fix the memory leakages in pingmsg.
* Set socket->retry to 5sec.

### DIFF
--- a/src/mqtt/transport/tcp/mqtt_tcp.c
+++ b/src/mqtt/transport/tcp/mqtt_tcp.c
@@ -776,7 +776,9 @@ mqtt_tcptran_pipe_recv_cb(void *arg)
 			property_free(prop);
 		}
 		// aio_begin?
-		if (!nni_aio_busy(p->qsaio)) {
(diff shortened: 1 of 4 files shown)


src/mqtt/transport/tcp/mqtt_tcp.c
```diff
@@ -786,10 +788,10 @@ mqtt_tcptran_pipe_recv_cb(void *arg)
 			// // send ACK down...
 			// nni_aio_set_iov(p->qsaio, 2, iov);
 			// nng_stream_send(p->conn, p->qsaio);
-		} else {
+		// } else {
 			// let protocol layer handle ack msg for us
```
wanghaEMQ, line 792: Protocol layer can handle ack msg. right? @JaylinYu

JaylinYu (commented):
thats luky I figure it out in the last min

PR #897, opened 2024-03-21
Sync transport & protocol layer changes that related to ping&qos

wanghaEMQ:


JaylinYu (approved):
Code is not the problem..... commit history is

JaylinYu:
some tweaks still needed.

PR #899, opened 2024-03-22
* FIX [parquet] When initializing the parquet file queue, remove any …

RanMaoyi:
…parquet files that have not yet been written

### DIFF
--- a/src/supplemental/nanolib/parquet/parquet.cc
+++ b/src/supplemental/nanolib/parquet/parquet.cc
@@ -680,6 +680,20 @@ parquet_file_queue_init(conf_parquet *conf)
 					closedir(dir);
 					return;
 				}
+				sprintf(file_path, "%s/%s", conf->dir,
+				    ent->d_name);
+
+				if (strstr(ent->d_name, "_") == NULL) {
+					if (unlink(file_path) != 0) {
+						log_error("Failed to remove file %s errno: %d",
+						    file_path, errno);
+					} else {
+						log_warn("Found a file without md5sum, "
+					         "delete it: %s", file_path);
+					}
+					free(file_path);
+					continue;
+				}
 
 				if (++count > conf->file_count) {
 					if (0 == remove_old_file()) {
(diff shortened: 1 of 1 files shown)


src/supplemental/nanolib/parquet/parquet.cc
```diff
@@ -688,11 +702,10 @@ parquet_file_queue_init(conf_parquet *conf)
 						         "file count: %d.",
 						    conf->file_count);
 					} else {
+						free(file_path);
```
xinyi-xs, line 705: This could result in a double free situation.

xinyi-xs (changes requested):

src/supplemental/nanolib/parquet/parquet.cc
```diff
@@ -680,6 +680,20 @@ parquet_file_queue_init(conf_parquet *conf)
...
 				}
+				sprintf(file_path, "%s/%s", conf->dir,
+				    ent->d_name);
+
+				if (strstr(ent->d_name, "_") == NULL) {
+					if (unlink(file_path) != 0) {
+						log_error("Failed to remove file %s errno: %d",
+						    file_path, errno);
+					} else {
+						log_warn("Found a file without md5sum, "
```
wanghaEMQ, line 691: Why there will be a file without md5sum?

src/supplemental/nanolib/parquet/parquet.cc
RanMaoyi: don't understand, file_path not enqueue parquet_file_queue, where will cause double free?

src/supplemental/nanolib/parquet/parquet.cc
RanMaoyi: parquet writes to a file by first creating a file without md5 and then renaming it to a file with md5 after the write is complete, so if nanomq quits while the file is being written, the file without md5 will be produced.

src/supplemental/nanolib/parquet/parquet.cc
xinyi-xs: See function remove_old_file(void).

RanMaoyi: But this file_path was created by alloc and is not queued yet. The file_path in remove_old_file() is taken from the queue.

PR #903, opened 2024-03-26
fix for https://github.com/nanomq/nanomq/issues/1719

OdyWayne:


src/sp/protocol/mqtt/nmq_mqtt.c
OdyWayne, line 1: line 621 `conn_param` has been used to init `p->keepalive`, but it's freed in line 612. So it looks like a u-a-f, better init it in another way.

wanghaEMQ: Does ASAN report it? It's also possible that cparam has been cloned before nni_pipe_get_conn_param.

wanghaEMQ (approved):

src/sp/protocol/mqtt/nmq_mqtt.c
OdyWayne: No, it's from a test report from the community. You are right about the clone thing. I think it's fine.

JaylinYu (approved):

PR #906, opened 2024-03-28
fix h-b-o #1720

OdyWayne:


OdyWayne:
not quite sure about this, plz review this.

src/supplemental/nanolib/hocon.c
```diff
@@ -66,7 +66,9 @@ path_expression_parse(cJSON *jso)
 			path_expression_parse(child);
 		}
 		if (NULL != child->string &&
-		    NULL != strchr(child->string, '.')) {
+		    NULL != strchr(child->string, '.') &&
+		    (parent->string != NULL &&
+		        strncmp(parent->string, "password", 8))) {
```
OdyWayne, line 71: this may cause potential issue when extend the usage of conf, especially when using `password` as parent .
Or we could just suggest users to avoid using dot in items.

src/sp/protocol/mqtt/mqtt_parser.c
```diff
@@ -583,16 +583,16 @@ conn_handler(uint8_t *packet, conn_param *cparam, size_t max)
...
 	cparam->keepalive_mqtt = tmp;
 	pos += 2;
-	// properties
 
+	// properties
 	if (cparam->pro_ver == MQTT_PROTOCOL_VERSION_v5) {
 		// check length
 		log_trace("Decoding MQTT V5 Properties");
-		if (pos >= max)
+		if (pos + 4 >= max)
```
wanghaEMQ, line 591: 4 == 2 (property len) + 2 (clientid len). Am I right?

src/supplemental/nanolib/conf_ver2.c
```diff
@@ -628,6 +628,13 @@ conf_auth_parse_ver2(conf *config, cJSON *jso)
 			    auth->usernames, nng_strdup(ele->string));
 			cvector_push_back(
 			    auth->passwords, nng_strdup(ele->valuestring));
+		} else if(cJSON_IsNumber(ele)) {
+			cvector_push_back(
+			    auth->usernames, nng_strdup(ele->string));
+			// max len of pw is 20.
+			char strpw[20];
+			sprintf(strpw, "%d", ele->valueint);
+			cvector_push_back(auth->passwords, nng_strdup(strpw));
```
wanghaEMQ, line 637: I don't think the password can be a number

wanghaEMQ (approved):

src/sp/protocol/mqtt/mqtt_parser.c
OdyWayne: it's the req of `get_var_integer`.

src/supplemental/nanolib/conf_ver2.c
OdyWayne: just revert it, gonna update docs and tell users to add `""` .

JaylinYu (approved):

PR #910, opened 2024-04-12
Add dot in key string support. 

xinyi-xs:


JaylinYu (approved):
This is what we long for

PR #925, opened 2024-04-22
* FIX [protocol/mqtt] fix https://github.com/nanomq/nanomq/issues/1762

JaylinYu:


wanghaEMQ (approved):
Coooooooool! That's the problem.

xinyi-xs (approved):

PR #960, opened 2024-05-21
Close tcp connection when received repeat mqtt connect msgs.

wanghaEMQ:
* Close tcp connection when received repeat mqtt connect msgs.


xinyi-xs (approved):

JaylinYu:
need to review the mqtt ws transport

JaylinYu (approved):

PR #974, opened 2024-05-31
Sync new fix from upstream, fix minor erros in sub_async api

JaylinYu:


xinyi-xs (approved):

JaylinYu:
apparently garrett is having trouble with win based nng

PR #977, opened 2024-06-03
Fix broker aio already in list.

wanghaEMQ:


xinyi-xs (approved):

JaylinYu (changes requested):
Aa We discussed, still has issue remain

xinyi-xs (approved):

JaylinYu (approved):
not fixed yet

PR #1004, opened 2024-07-04
Sync to updated 0.22.1

JaylinYu:


include/nng/supplemental/nanolib/conf.h
```diff
@@ -315,6 +315,7 @@ struct conf_exchange_node {
 
 	nng_socket       *sock;
 	nng_mtx          *mtx;
+	char             *exchange_url;
```
JaylinYu, line 318: Just realized this.
I prefer to settle multi-stream query in self-defined protocol. instead of  assigning unique URL to each stream.
@RanMaoyi

JaylinYu (commented):
@RanMaoyi plz review multi-stream sync part
@xinyi-xs plz review new uds logging part

wanghaEMQ (approved):

include/nng/supplemental/nanolib/conf.h
RanMaoyi: Just thinking about it feels like a huge project, with a lot of decoupling to do, so we can plan it out later.

RanMaoyi (approved):

PR #1005, opened 2024-07-05
* FIX [mqtt_codec] variable length is only 4 bytes max

JaylinYu:
Fix https://github.com/nanomq/nanomq/issues/1810


wanghaEMQ (commented):
Good

PR #1038, opened 2024-07-31
* MDF [mqtt_msg] port new fix to sdv

JaylinYu:


JaylinYu:
hotfix from nanomq master 0.22.4

PR #1042, opened 2024-08-06
Fix new coming bugs!

JaylinYu:
https://github.com/nanomq/nanomq/issues/1863


wanghaEMQ (commented):
Need more benchmark tests

JaylinYu:
> Need more benchmark tests

indeed, there is definitely performance degradation.

PR #1048, opened 2024-08-12
Fix of https://github.com/nanomq/nanomq/issues/1861

JaylinYu:


src/supplemental/mqtt/mqtt_codec.c
```diff
@@ -455,17 +456,17 @@ nni_mqtt_msg_dup(void **dest, const void *src)
 			dup_connect(mqtt, s);
 		}
 		if (s->var_header.connect.properties) {
-			property_dup(&mqtt->var_header.connect.properties,
+			rv += property_dup(&mqtt->var_header.connect.properties,
```
wanghaEMQ, line 459: Bad definition of the rv :<

JaylinYu: why?

JaylinYu: why?

PR #1049, opened 2024-08-12
Add hybrid_servers to bridge config.

wanghaEMQ:
* Add hybrid_servers to bridge config.
* hybrid_bridging is no longer supported in old conf.


JaylinYu (approved):
Can I only set tcp server in hybrid list?

xinyi-xs (approved):

PR #1054, opened 2024-08-21
Paruqet new schema(write).

xinyi-xs:


src/supplemental/nanolib/parquet/parquet.cc
```diff
@@ -168,38 +195,87 @@ parquet_file_range_free(parquet_file_range *range)
...
+parquet_data_free(parquet_data *data)
+{
+	if (data) {
+		for (uint32_t c = 0; c < data->col_len-1; c++) {
+			FREE_IF_NOT_NULL(
+			    data->schema[c], strlen(data->schema[c]));
+			for (uint32_t r = 0; r < data->row_len; r++) {
+				parquet_data_packet *payload =
+				    data->payload_arr[c][r];
+				if (payload) {
```
RanMaoyi, line 229: remove this line, null will check in FREE_IF_NOT_NULL

JaylinYu (approved):
plz leave comments so late comer could understand u

RanMaoyi (approved):

PR #1055, opened 2024-08-21
Add libargon for encrypt and implement password check

StargazerWayne:


JaylinYu (commented):
Surprise!

JaylinYu:
include a encry lib via binary is not legit. 
Will replace it with submodule or cmake  find library

PR #1079, opened 2024-10-12
* FIX [mqtt_codec] var max pos is 5 instead of 4

RanMaoyi:


xinyi-xs (approved):

JaylinYu:
Seems like I didnt fix it right

JaylinYu (approved):

PR #1104, opened 2024-11-01
Reopen quic sub stream actively when send timeout.

wanghaEMQ:


src/supplemental/quic/msquic_dial.c
```diff
@@ -214,6 +216,33 @@ verify_peer_cert_tls(QUIC_CERTIFICATE* cert, QUIC_CERTIFICATE* chain, char *ca)
...
+				break;
+			}
+			nni_mtx_unlock(&c->mtx);
+			c = NULL;
+		}
+	}
+	nni_mtx_unlock(&ec->mtx);
+
+	if (c)
+		quic_substream_close(c);
```
JaylinYu, line 242: If so, isnt it always only 1 sub stream will be closed?

JaylinYu, line 242: only low priority stream needs this

JaylinYu (changes requested):

src/supplemental/quic/msquic_dial.c
wanghaEMQ: Yes. In a round. There is only one sub stream will be closed. If another sub stream also sends timeout. It will be closed at next round.

wanghaEMQ: Yes. Only low priority streams has update of timeout. For the main stream and high priority streams. The timeout are always 0.

JaylinYu (approved):
toooooo much locks

PR #1106, opened 2024-11-08
* FIX [stream] RAW stream need to set schema by default

RanMaoyi:


JaylinYu (approved):
kuiper need to have a try

xinyi-xs (approved):

PR #1117, opened 2024-11-20
Add level pid file line func info to syslog

xinyi-xs:


JaylinYu (approved):
main wants it as well

wanghaEMQ (commented):
Cool

PR #1122, opened 2024-11-28
Multiple enhancements

JaylinYu:


wanghaEMQ (commented):
Cool. Are there any documents about `BUILD_STATIC` need to be updated?

PR #1157, opened 2025-03-03
Fixed parquet add queue error.

xinyi-xs:


wanghaEMQ (approved):

src/supplemental/nanolib/parquet/parquet_file_manager.cc
```diff
@@ -4,8 +4,8 @@
 void
 parquet_file_manager::add_queue(conf_exchange_node *node)
 {
-	file_queue_map[node->name] = make_shared<parquet_file_queue>(node->parquet);
-	file_queue_map[node->name]->init();
+	file_queue_map[node->topic] = make_shared<parquet_file_queue>(node->parquet);
```
JaylinYu, line 7: Need a urgent release?

JaylinYu (approved):

src/supplemental/nanolib/parquet/parquet_file_manager.cc
xinyi-xs: If config exchange name is the same with topic, it's OK and normaly it's same.

PR #1183, opened 2025-04-11
Support to read aes encrypted certs/ca/key.

wanghaEMQ:


xinyi-xs (approved):

JaylinYu (commented):
too much redundant changes.
how about put them at file_reload data together?

wanghaEMQ:
> too much redundant changes. how about put them at file_reload data together?

Yes. Done.

JaylinYu:
although not my original thoughts, but looks decent now

PR #1189, opened 2025-04-15
* FIX [mqtt_quic] cached aio with subscription packet turn pipe busy

JaylinYu:


JaylinYu:
fix a sidling issue spotted with https://github.com/nanomq/nanomq/issues/2011

PR #1206, opened 2025-05-08
preset session feature

JaylinYu:


JaylinYu:
resuming seesion still needs to subscribe to topics by itself
preset sessions with SQLite enabled is not tested yet

wanghaEMQ (approved):

PR #1210, opened 2025-05-14
New apis to get retain msgs

wanghaEMQ:


wanghaEMQ:
> can we just reads retain msg from dbtree??

Yes. Now we can

JaylinYu (approved):
not compatible with SQLite

PR #1227, opened 2025-06-05
feat: add property retry_qos_0

ftapajos:
It was my first time exploring this codebase, please be kind.

I need to disable caching of qos 0 mqtt messages for bridges when the bridge is not available. [My] applications are usually developed under the assumption that QoS0 messages don't have the guarantee of being delivered, hence they should not be persisted if the connection is not available. 

When the connection is available available most of the time, this should be a minor problem, but for our case brokers are placed in remote networks that may have no internet access for several days until proper maintenance is performed. In such cases, qos0 messages are unimportant and disk space should be left to store qos1 and qos2 messages.


JaylinYu:
Thanks for your proposal, gonna add a new config param for this.

PR #1270, opened 2025-07-24
Add Raft conf parsing

Elia-Renzoni:
I closed the previous PR because it was minimal, and I wanted to do something more complete.
In this PR, I’ve added further changes and improvements compared to the previous version.

Let me know what you think!


JaylinYu:
I would propose a new nano-raft branch for exploring the clustering feature. this is not sth could be chopped into pieces as small PR.

Elia-Renzoni:
Hi @JaylinYu  
the clustering feature will need to be implemented in NanoMQ, and I’ll follow your suggestion to work locally and only open a PR once everything is complete.

The idea behind this PR was to already have the necessary raft-related data structures merged into the NanoNNG submodule, so that part would be in place.
However, I can definitely work on a single PR that includes everything instead.

PR #1310, opened 2025-08-19
base64 decode and then decrypt the password in http_auth

wanghaEMQ:


xinyi-xs (approved):

RanMaoyi:
Also, it's good practice to check for NULL after alloc.

PR #1336, opened 2025-09-05
New field exec_fname, nanomq could have different name.

wanghaEMQ:


JaylinYu (commented):
Do We really need 64bytes long？

PR #1351, opened 2025-09-18
* FIX [conf] Fix the error in compiling.

wanghaEMQ:


src/supplemental/nanolib/conf_ver2.c
```diff
@@ -15,7 +15,7 @@
 #include <ctype.h>
 #include <string.h>
 
-#ifdef SUPP_LICENSE_STD
+#if defined(SUPP_LICENSE_STD) || defined(SUPP_PARQUET)
 #include "supplemental/aes/aes.h"
 #include "supplemental/base64/base64.h"
```
JaylinYu, line 20: same behaviour in rest_api , you sure?

JaylinYu: 750 + 2251line as well
plz check again

JaylinYu (changes requested):

PR #1363, opened 2025-10-09
Add a unique ID to the Parquet filename to support rolling updates

xinyi-xs:


src/supplemental/nanolib/parquet/parquet_file_queue.cc
```diff
@@ -52,34 +52,78 @@ parquet_file_queue::init()
...
+					    extract_start_time(file_name);
+
+					// If start time extraction fails, skip
+					// the file
+					if (!start_time.has_value()) {
+						log_error(
+						    "Failed to extract start "
+						    "time from file: %s",
+						    file_name.c_str());
+						continue;
```
JaylinYu: delete?

PR #1392, opened 2025-12-24
Fix rule_sql_parse from-clause handling logic.

xinyi-xs:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL query parsing to correctly detect and extract FROM clauses only when long enough, preventing incorrect slicing/underflow and improving stability when handling short or malformed queries.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


JaylinYu:
please also look at rabbit' review

PR #1469, opened 2026-03-26
Fix: Add try_private option to control connection bridge bit

tricera-josua-palmstedt:
### Problem

The MQTT CONNECT encoder in `mqtt_codec.c` hardcoded `0x84` as the protocol version byte when built with `NNG_HAVE_MQTT_BROKER`. This unconditionally set the Mosquitto bridge bit (`0x80`) on every bridge CONNECT, making it impossible to connect to MQTT brokers that do not support this non-standard extension.

### Fix

1. **Remove the `#ifdef` hack from the CONNECT encoder** (`mqtt_codec.c`): The encoder now always writes `var_header->protocol_version` as-is. The caller is responsible for setting the bridge bit if needed.

2. **Add `try_private` field to `conf_bridge_node`** (`conf.h`): Boolean flag, defaults to `true` for backward compatibility.

3. **Parse `try_private` from config** (`conf.c`, `conf_ver2.c`): Supported in v2 (HOCON) config format, and included in `print_bridge_conf` output.

### Changes

| File | Change |
|------|--------|
| `src/supplemental/mqtt/mqtt_codec.c` | Remove `#ifdef NNG_HAVE_MQTT_BROKER` block, always use `var_header->protocol_version` |
| `include/nng/supplemental/nanolib/conf.h` | Add `bool try_private` to `conf_bridge_node` |
| `src/supplemental/nanolib/conf.c` | Default `try_private = true`, add v1 config parsing, add to `print_bridge_conf` |
| `src/supplemental/nanolib/conf_ver2.c` | Add v2 (HOCON) config parsing for `try_private` |

## Summary by CodeRabbit

* **New Features**
  * Added a per-bridge-node boolean configuration option "try_private", enabled by default, exposed in configuration parsing and shown in bridge logging.

* **Refactoring**
  * MQTT CONNECT protocol version is now derived from configured values consistently, simplifying protocol-version handling.

### DIFF
--- a/src/supplemental/mqtt/mqtt_codec.c
+++ b/src/supplemental/mqtt/mqtt_codec.c
@@ -856,11 +856,7 @@ nni_mqtt_msg_encode_connect(nni_msg *msg)
 	nni_mqtt_msg_encode_fixed_header(msg, mqtt);
 
 	nni_mqtt_msg_append_byte_str(msg, &var_header->protocol_name);
-#ifdef NNG_HAVE_MQTT_BROKER
-	nni_mqtt_msg_append_u8(msg, 0x84);	// hack first bit as bridge flag
-#else
 	nni_mqtt_msg_append_u8(msg, var_header->protocol_version);
-#endif
 
 	/* Connect Flags */
 	nni_mqtt_msg_append_u8(msg, *(uint8_t *) &var_header->conn_flags);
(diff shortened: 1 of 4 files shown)


src/supplemental/mqtt/mqtt_codec.c
```diff
@@ -856,11 +856,7 @@ nni_mqtt_msg_encode_connect(nni_msg *msg)
 	nni_mqtt_msg_encode_fixed_header(msg, mqtt);
 
 	nni_mqtt_msg_append_byte_str(msg, &var_header->protocol_name);
-#ifdef NNG_HAVE_MQTT_BROKER
```
JaylinYu, line 859: do not remove the MACRO conditon

JaylinYu (approved):

PR #1488, opened 2026-04-08
Each exchange has its own encryption options

wanghaEMQ:
## Summary by CodeRabbit

* **Refactor**
  * Reorganized encryption configuration handling to embed settings within individual nodes rather than using shared pointers, improving memory management and structural consistency.
  * Updated OpenSSL API calls with explicit type casting for enhanced compatibility.


wanghaEMQ:
@coderabbitai review

PR #1490, opened 2026-04-14
Fix bridge client missing config auth mode

alvin1221:
Fix bridge client missing config auth mode

> Add `tls.verify_peer` = `true` or `false` to bridge tls config is necessary


JaylinYu (approved):
only reload>??

alvin1221:
> only reload>??

Yes， you can try with config `verify_peer` = false /  true and use invalid certification file or miss-match Server Name Indication

JaylinYu:
It shall be a standard config param that taking effect in init_dialer_tls

JaylinYu:
it actually works in sdv mode. because we call tls_reolad even first time connect. however it doesnt work in open source ver.

PR #1500, opened 2026-04-23
Fix a heap use after free bug of conn_param

JaylinYu:
## Bug Fixes
* Improved MQTT message handling with refined connection parameter cleanup logic, optimizing resource usage for specific message types
* Enhanced error handling for scenarios where messages arrive without a receiver context, improving system resilience and preventing unscheduled message handling issues
* Optimized transport layer cleanup behavior for better resource management

### DIFF
--- a/src/sp/protocol/mqtt/nmq_mqtt.c
+++ b/src/sp/protocol/mqtt/nmq_mqtt.c
@@ -843,11 +843,11 @@ close_pipe(nano_pipe *p)
 
 	nni_atomic_set_bool(&p->closed, true);
 	if (nni_list_active(&s->recvpipes, p)) {
-		nni_msg *msg = nni_aio_get_msg(&p->aio_recv);
-					uint8_t type = nng_msg_cmd_type(msg);
-	if (type == CMD_SUBSCRIBE || type == CMD_UNSUBSCRIBE ||
-		   		type == CMD_CONNACK || type == CMD_PUBLISH)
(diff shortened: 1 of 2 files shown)


JaylinYu:
@coderabbitai review

src/sp/protocol/mqtt/nmq_mqtt.c
```diff
@@ -843,10 +843,16 @@ close_pipe(nano_pipe *p)
 
 	nni_atomic_set_bool(&p->closed, true);
 	if (nni_list_active(&s->recvpipes, p)) {
-		nni_msg *msg = nni_aio_get_msg(&p->aio_recv);
-		if (msg)
+		nni_msg *msg  = nni_aio_get_msg(&p->aio_recv);
+		if (msg) {
+			uint8_t  type = nng_msg_cmd_type(msg);
+			if (type == CMD_SUBSCRIBE || type == CMD_UNSUBSCRIBE ||
+				type == CMD_CONNACK || type == CMD_PUBLISH)
```
wanghaEMQ, line 850: Why don't you do a conn_param_clone for all kinds of type. so you don't need to check the type and do free

JaylinYu: historical reason
you can change it if you want

wanghaEMQ: fine

PR #1512, opened 2026-05-06
* FIX [quic] Fixed the type confusion when nni_quic_dialer_close.

wanghaEMQ:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QUIC connection cleanup handling to properly close associated resources and prevent potential connection leaks.


JaylinYu (approved):
doens look like a mistake can be made only because of careless

PR #1515, opened 2026-05-15
fix(http): don't abort broker on aio cleanup after HTTP timeout

lurtze1:
> Hi maintainers,
>
> While investigating a production-class stability problem in a NanoMQ
> deployment that uses the HTTP auth callback, we tracked the crash to
> two `exit(EXIT_FAILURE)` calls in `nni_http_conn_fini`. This PR removes
> them and falls through to the existing cleanup tail. The change is
> minimal and contained to one function in `src/supplemental/http/http_conn.c`.
>
> Detailed analysis below — happy to adjust the patch to whatever style you'd prefer.

I did get AI to help me on this a lot, wouldn't have figured it out myself but I am also not experienced enough in the software to say if it's conclusions are 100% correct.

### Summary

`nni_http_conn_fini` in `src/supplemental/http/http_conn.c` aborts the entire
broker process whenever `nni_aio_schedule` returns non-zero for the
connection's `wr_aio` or `rd_aio` during cleanup:

```c
if ((nni_aio_schedule(conn->wr_aio, nng_wr_fr_cb, conn)) != 0) {
    log_error("Non recoverable error, aio schedule failed!");
    exit(EXIT_FAILURE);
}
```

That return value, however, is the documented and expected result when
the aio is already in a terminal state (`a_abort = true`, `a_stop = true`)
— which is precisely when this finalizer is most likely to be invoked.
For example: any time `auth_http.c send_request` finalizes a conn after
hitting its HTTP timeout, the aios are already terminal and this exit
fires.

In production, the effect is that **one slow HTTP-auth round trip kills
the broker process**. All connected clients drop, reconnect en masse,
re-saturate the auth callback, and we cycle. The two `exit()` lines act
as a process-wide DoS amplifier under any sustained HTTP-auth load.

### Reproduction

Minimal reproduction we used:

1. NanoMQ 0.24.13 with `auth.http_auth` configured against a slow-ish
   HTTP endpoint (we used a Laravel webapp doing bcrypt(12) on each
   request, ~180 ms per check).
2. `http_auth.timeout = "2s"` (default, also our setting).
3. Open ~600+ concurrent MQTT clients each authenticating with unique
   credentials so the auth-callback queue saturates.

Result: the broker aborts within seconds with:

```
ERROR auth_http.c:316 send_request: Read response: Timed out
ERROR http_conn.c:721 nni_http_conn_fini: Non recoverable error, aio schedule failed!
```

…and Docker / systemd restarts the process. The cycle is fully reproducible.

### Root cause

`nni_aio_schedule` (in `src/core/aio.c`) returns non-zero in three documented
situations:

```c
if (aio->a_abort) {
    int rv = aio->a_result;
    nni_mtx_unlock(&eq->eq_mtx);
    return (rv);
}
if (aio->a_stop) {
    nni_mtx_unlock(&eq->eq_mtx);
    return (NNG_ECLOSED);
}
```

i.e. when the aio is aborted (typically: timed out) or stopped (typically:
the underlying conn is closing). Both of these are *normal* end-states
after an HTTP timeout or peer close. They are not invariant violations.

`nni_http_conn_fini` is invoked exactly on those code paths — after a
timeout, a write/read error, or a peer-initiated close. So the `exit()`
fires precisely when the aios are in their expected terminal state, not
on a real "non recoverable" error.

The bottom half of the function already handles the terminal case
correctly via `nni_aio_busy` + `nni_aio_free`/`nni_aio_reap`:

```c
if (!nni_aio_busy(conn->rd_aio) && !nni_aio_busy(conn->wr_aio)) {
    nni_aio_free(conn->wr_aio);
    nni_aio_free(conn->rd_aio);
} else {
    nni_aio_reap(conn->wr_aio);
    nni_aio_reap(conn->rd_aio);
}
```

The `exit()` calls simply interrupt that cleanup before it can run.

### The fix

Drop the two `exit(EXIT_FAILURE)` blocks and fall through to the
existing reap/free path. When scheduling the `nng_wr_fr_cb` /
`nng_rd_fr_cb` callback fails (because the aio is terminal), we mark
that side closed inline — those callbacks would otherwise do this for
us and also reap `fe_aio` once both sides are closed. Doing this
bookkeeping inline means `fe_aio` is still released when both schedule
calls fail.

```c
if (nni_aio_schedule(conn->wr_aio, nng_wr_fr_cb, conn) != 0) {
    conn->wr_close = true;
}
if (nni_aio_schedule(conn->rd_aio, nng_rd_fr_cb, conn) != 0) {
    conn->rd_close = true;
}
if (conn->wr_close && conn->rd_close && conn->fe_aio != NULL) {
    nni_aio_reap(conn->fe_aio);
    conn->fe_aio = NULL;
}
```

The change is +17 / -7 lines in one file. No API surface is touched.
No new control-flow path is added — we only stop interrupting the
existing one with a panic-exit.

### Testing

We ran three workloads against a patched broker (NanoMQ 0.24.13 with
this one change applied to its NanoNNG dependency, built from the
`Dockerfile-dev` flow):

| Test | Result before patch | Result after patch |
|---|---|---|
| Sequential add (1 device every ~180 ms, up to 1000) | Broker dies at ~660 devices, `aio schedule failed` | 0 occurrences of the original crash signature, broker stays up |
| Sequential add + auth cache (`cache_ttl = 5m`) | n/a | 0 broker restarts, 1 broker-side HTTP timeout, no crash |
| Reconnect storm (1000 simultaneous CONNECTs, with exponential reconnect backoff, 4 minutes sustained) | Broker dies within seconds | 0 broker restarts across 11 k+ reconnect attempts and 17 k+ log lines |

After the patch, `aio schedule failed!` log lines: **0 across all runs**.

### Known follow-ups (not in this PR)

While preparing this fix we noticed two related issues. They are not
required for this PR to be a strict improvement, but we'll happily
file separate issues / PRs if you want:

1. **Latent UAF in the HTTP cleanup path.** Once the `exit()` is removed
   the broker can reach further into the cleanup logic, where under
   high concurrency we have seen a `panic: pthread_mutex_lock: Invalid
   argument` / SIGABRT. The crash signature points to a use-after-free
   on a mutex-bearing struct in the HTTP teardown path
   (`nng_http_conn_close` → `nng_http_client_free`). In our workload
   we saw 0–2 of these per long run, and zero during a 4-minute
   reconnect storm; it's rare in practice. We can open a follow-up issue
   with the stack signature and a reproducer if useful.

2. **`auth_http.pool_size` is parsed but unused.** The configuration key
   is read into `conf_auth_http.pool_size` (`conf.c:4336`), logged at
   startup (`conf.c:1049`), but is never consumed by the auth path:
   `send_request` in `auth_http.c` allocates a fresh `nng_http_client`
   per request and serializes the whole roundtrip on `conf_req->mtx`,
   so all HTTP-auth calls are effectively single-threaded regardless
   of `pool_size`. Happy to open a separate issue for that.

We deliberately kept those out of this PR so the change stays small
and the bug fix can land independently of the bigger throughput work.

---

## What this PR is *not*

- It is not a behavior change for the happy path — successful auth
  calls finalize via the same `nni_aio_free` path they always did.
- It is not a fix for slow HTTP backends; the broker will still log
  `Read response: Timed out` when an HTTP request misses its deadline.
  It just no longer kills the process for it.
- It is not a complete cure for HTTP-auth crashes (see follow-up #1
  above). It is, however, a strict improvement: the old exit() path
  is gone, and the residual UAF is dramatically rarer under realistic
  workloads.

Thanks for the project — happy to iterate on the patch or split it
differently if you prefer.

### DIFF
--- a/src/supplemental/http/http_conn.c
+++ b/src/supplemental/http/http_conn.c
@@ -713,13 +713,23 @@ void
 nni_http_conn_fini(nni_http_conn *conn)
 {
 	nni_mtx_lock(&conn->mtx);
-	if ((nni_aio_schedule(conn->wr_aio, nng_wr_fr_cb, conn)) != 0) {
-		log_error("Non recoverable error, aio schedule failed!");
-		exit(EXIT_FAILURE);
+	// nni_aio_schedule returns non-zero when the aio is already in a
+	// terminal state (aborted/timed-out/closed). That is the *expected*
+	// situation when fini runs after a timeout or peer-close — not a
+	// fatal error. Previously this aborted the whole broker process,
+	// which under sustained HTTP-auth load is a denial-of-service.
+	// If scheduling fails, the fr_cb callback will not run, so we
+	// mark that side closed inline and reap fe_aio ourselves once both
+	// sides have bypassed the callback path.
+	if (nni_aio_schedule(conn->wr_aio, nng_wr_fr_cb, conn) != 0) {
+		conn->wr_close = true;
 	}
-	if ((nni_aio_schedule(conn->rd_aio, nng_rd_fr_cb, conn)) != 0) {
-		log_error("Non recoverable error, aio schedule failed!");
-		exit(EXIT_FAILURE);
+	if (nni_aio_schedule(conn->rd_aio, nng_rd_fr_cb, conn) != 0) {
+		conn->rd_close = true;
+	}
+	if (conn->wr_close && conn->rd_close && conn->fe_aio != NULL) {
+		nni_aio_reap(conn->fe_aio);
+		conn->fe_aio = NULL;
 	}
 	conn->free = true;
 	http_close(conn);


JaylinYu:
May I know what AI model you are using for this PR?
And how is it working in your application with this PR?

lurtze1:
I was using claude Opus 4.7. 
I was trying to figure out why my connections were dropping for nanomq and the container was restarting when i tried to create a renew storm on a authentication http endpoint on my webserver. So nanomq has an auth hook to my webserver that linked to a endpoint that allowed me to check the username and password of the connecting client and compare it to a record in my database.

There are more underlying issues, I got mutex errors after more testing with this "fix" but I had to transition to using mTLS as authentication anyway so I pivoted to using mosquitto and mTLS.

JaylinYu:
I know, I replied you via slack
Ok, I will look into it and make it good enough to be merged. 
At least make your expensive token count. even Opus has limits.

alvin1221 (approved):

alvin1221:
@JaylinYu  I have verified that this PR is approved. Below is the comparison of the test report before and after the modifications.
[pr1515-validation-en.md](https://github.com/user-attachments/files/28091253/pr1515-validation-en.md)

JaylinYu:
> @JaylinYu I have verified that this PR is approved. Below is the comparison of the test report before and after the modifications. [pr1515-validation-en.md](https://github.com/user-attachments/files/28091253/pr1515-validation-en.md)

Generated & tested by what?

alvin1221:
> > @JaylinYu I have verified that this PR is approved. Below is the comparison of the test report before and after the modifications. [pr1515-validation-en.md](https://github.com/user-attachments/files/28091253/pr1515-validation-en.md)
> 
> Generated & tested by what?



Here'are the test script and log;
[http_auth_pr1515_tests.py](https://github.com/user-attachments/files/28100434/http_auth_pr1515_tests.py)
[http_auth_mock_server.py](https://github.com/user-attachments/files/28100435/http_auth_mock_server.py)
[run_http_auth_pr1515_tests.sh](https://github.com/user-attachments/files/28100436/run_http_auth_pr1515_tests.sh)
[log.zip](https://github.com/user-attachments/files/28100393/log.zip)

Test shell: 
```bash
./run_http_auth_pr1515_tests.sh --scenario all --profile full --out-dir logs/pr1515_full_$(date +%s)
```

alvin1221:
> I was using claude Opus 4.7. I was trying to figure out why my connections were dropping for nanomq and the container was restarting when i tried to create a renew storm on a authentication http endpoint on my webserver. So nanomq has an auth hook to my webserver that linked to a endpoint that allowed me to check the username and password of the connecting client and compare it to a record in my database.
> 
> There are more underlying issues, I got mutex errors after more testing with this "fix" but I had to transition to using mTLS as authentication anyway so I pivoted to using mosquitto and mTLS.


I've reproduced this issue, here's the log file [broker.log](https://github.com/user-attachments/files/28133876/broker.log).

JaylinYu:
We got another fix PR regarding this issue. https://github.com/nanomq/NanoNNG/pull/1578/changes.
Perhaps your Opus 4.7 is just one step distance from fixing it. Opus 4.7 doesn't understand the aio_reap is the finit action. That's why you got a mutex finit crash.

@lurtze1 would be appreciated if you could try again in your case to verify the fix

JaylinYu:
> > I was using claude Opus 4.7. I was trying to figure out why my connections were dropping for nanomq and the container was restarting when i tried to create a renew storm on a authentication http endpoint on my webserver. So nanomq has an auth hook to my webserver that linked to a endpoint that allowed me to check the username and password of the connecting client and compare it to a record in my database.
> > There are more underlying issues, I got mutex errors after more testing with this "fix" but I had to transition to using mTLS as authentication anyway so I pivoted to using mosquitto and mTLS.
> 
> I've reproduced this issue, here's the log file [broker.log](https://github.com/user-attachments/files/28133876/broker.log).

You are not being helpful

PR #1549, opened 2026-06-17
upgrade to hash64 of pipeid

JaylinYu:
## Summary by CodeRabbit

* **Improvements**
  * MQTT identifier handling is now 64-bit across sessions, pipes, and related databases.
  * Hashing for MQTT client identifiers uses a 64-bit FNV-1a based key for better distribution.
  * Statistics and scope reporting now support 64-bit identifier values.
* **Bug Fixes**
  * Correctly reads the MQTT V5 “Maximum Packet Size” property when setting the client’s max packet size.
* **Compatibility**
  * Updated public MQTT/QoS database APIs and pipe/message “pipe” identifiers to use 64-bit types.

### DIFF
--- a/include/nng/mqtt/packet.h
+++ b/include/nng/mqtt/packet.h
@@ -47,7 +47,7 @@
 	} while (0)
 
 struct mqtt_msg_info {
-	uint32_t pipe;
+	uint64_t pipe;
 };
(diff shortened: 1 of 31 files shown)


JaylinYu:
@coderabbitai review

JaylinYu:
failure on trying

PR #1557, opened 2026-06-29
Add android keystore2 support

alvin1221:
nanomq PR: https://github.com/emqx/NanoMQ_mirror/pull/620


alvin1221:
@coderabbitai review

wanghaEMQ (changes requested):
What a hell. There must be a better way to import these blinder file. Add them to supplemental directly looks bad

alvin1221:
> What a hell. There must be a better way to import these blinder file. Add them to supplemental directly looks bad

I'm thinking of moving them to NDK;

wanghaEMQ:
Should all remaining files be put in the supplemental folder? for example, blinder.so... And I think it will be better to create a file named tee3.c and move most of the changes in openssl.c to tee3.c

PR #1560, opened 2026-07-04
* FIX [broker] redeliver offline QoS msgs to resumed persistent sessions

ingwinlu:
## Summary

A persistent session (`clean_start=false`) that reconnects after a brief outage now catches up on its offline QoS-1/2 backlog. Before this change, with SQLite persistence enabled:

- a **shared subscription** (`$share/<group>/<filter>`) received **none** of the messages published while it was offline — the transport cache branch re-matched the publish topic against the raw stored filter, the `$share` prefix never matched, and the message was freed instead of stored;
- a **plain subscription** received its backlog at **one message per `retry_interval`** — redelivery was driven only by the resend timer, so a 5-message backlog took 5+ intervals to drain;
- for roughly **half of all client ids** (siphash above `INT32_MAX`), persisted messages could be neither fetched nor ACK-removed at all: `get_one`/`remove`/`remove_by_pipe` bound the `uint32` pipe id with 32-bit `sqlite3_bind_int` (sign-extended negative) while the writers bind 64-bit.

After this change, the backlog is stored for both subscription types and drained right after the CONNACK at ACK round-trip rate (5 messages in ~0.1s in the regression, vs ~25s+ before), on all three broker transports and both persistence backends.

## Design

- **Store fix:** a `shared_filter_skip()` helper strips `$share/<group>/` before `topic_filtern` in the offline cache branch of tcp/tls/ws — the same strip every online send path already does. The branch now also keeps the strongest matching subscription instead of first-match-wins, so an overlapping QoS0 filter cannot shadow a QoS1 one.
- **ACK-clocked drain:** resuming a cached session arms a drain; dispatching the CONNACK starts it (CONNACK-before-PUBLISH stays causal per MQTT-3.2.0-1), and each terminal ACK of the drained message chains the next fetch after the transport removed the acked row. Drain fetches bypass the retry age gate via a new `nmq_req.drain` flag. The drain disarms when the store is empty or the head row is a live in-flight message (stored after resume), and hands over to the resend timer — which remains the retry safety net, unchanged.
- **Boundary hardening** (from review): the drain is disarmed when a session is parked into the cache mid-drain (a pending fetch on a cached pipe would re-store the message as a duplicate row); only the drained message's own ack clocks the next fetch (an ack of interleaved live traffic must not re-fetch an un-PUBCOMP'd QoS2 head after PUBREL); the websocket transport now timestamps received PUBLISHes like tcp/tls (ws-published messages carried ts=0 and were misclassified as backlog).
- **Ownership fix:** SQLite's `get_one` returns a freshly deserialized, solely-owned message; the resend getopt no longer clones it (one whole message leaked per resend) and frees it when not returned.
- **Deterministic FIFO:** `get_one` gained `ORDER BY main.id`, so SQLite redelivery follows publish order.
- Sessions restored **across a broker restart** keep timer-paced redelivery: message timestamps are monotonic per-process clocks and are not comparable between processes, so the fast drain's backlog/in-flight distinction cannot be trusted there.

## Validation

Unit tests cover the strip helper (incl. malformed `$share` shapes), FIFO order, pipe-rebind and removal with pipe ids above `INT32_MAX`. The companion nanomq PR adds an end-to-end regression driving this code: {plain, `$share`} x {SQLite, in-memory} redelivery in ~0.1s with exactly-once and strict-FIFO assertions, QoS-0 controls, live publishes interleaved into a stalled drain (manual acks), and SQLite persistence across a broker restart — all green repeatedly under an ASAN build.

Related: nanomq/nanomq#2355

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added an ACK-paced session-resume backlog drain that reliably continues delivering stored QoS messages after reconnects, with controlled retries and ordering safeguards.
* **Bug Fixes**
  * Shared-subscription matching now always uses the strongest effective QoS across all matching filters (QoS0 no longer masks higher QoS).
  * QoS resend/restore handling now correctly interprets cached entries for drain mode and removes expired QoS messages per configured duration.
* **Tests**
  * Added QoS database tests for FIFO retrieval, pipe-id rebind after restart, and pipe-based removal cleanup.

### DIFF
--- a/src/sp/protocol/mqtt/nmq_mqtt.c
+++ b/src/sp/protocol/mqtt/nmq_mqtt.c
@@ -25,6 +25,10 @@
 
 #define DB_NAME "nano_qos_db.db"
 
+// session-resume backlog drain: kickoff retry budget and backoff (ms)
+#define NANO_DRAIN_RETRY 10
+#define NANO_DRAIN_INTERVAL 100
+
 typedef struct nano_pipe   nano_pipe;
 typedef struct nano_sock   nano_sock;
 typedef struct nano_ctx    nano_ctx;
@@ -32,6 +36,7 @@ typedef struct cs_msg_list cs_msg_list;
 
 static void        nano_pipe_send_cb(void *);
 static void        nano_pipe_recv_cb(void *);
+static void        nano_pipe_drain_cb(void *);
 static void        nano_pipe_fini(void *);
 static int         nano_pipe_close(void *);
 static inline void close_pipe(nano_pipe *p);
@@ -85,11 +90,29 @@ struct nano_pipe {
 	                     	// timer has been triggered
 	bool          busy;
 	bool          event; // indicates if exposure disconnect event is valid
+	bool          resumed;  // session resumed; drain armed once CONNACK
+	                        // enters the send path
+	bool          draining; // backlog drain in progress: ACK-clocked
+	                        // fetches of stored QoS msgs (age gate bypassed)
+	bool          drain_fetch_due; // one-shot: fetch when pipe goes idle
+	                               // (set at CONNACK dispatch and on ACK
+	                               // arriving while a send is in flight)
+	bool          drain_sent;  // drain delivered at least one msg
+	bool          drain_wait;  // aio_drain retry sleep is armed
+	uint8_t       drain_retry; // remaining kickoff retries
+	uint16_t      drain_pid;   // packet id of the in-flight drained msg
+	                           // (0 when none): only its own terminal ack
+	                           // chains the next fetch, so acks of
+	                           // interleaved live traffic cannot re-fetch
+	                           // an un-PUBCOMP'd QoS2 head
+	nni_time      resume_time; // when the session resumed; stored msgs newer
+	                           // than this are in-flight, not backlog
 	void         *tree;  // root node of db tree
 	void         *nano_qos_db; // 'sqlite' or 'nni_id_hash_map'
 	nni_aio       aio_send;
 	nni_aio       aio_recv;
 	nni_aio       aio_timer;
+	nni_aio       aio_drain; // backlog-drain retry timer
 	nni_list_node rnode; // receivable list linkage
 	nni_atomic_bool closed;
 };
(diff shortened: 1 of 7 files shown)


JaylinYu:
I just recall that We had a similiar PR before....
but your new one still stands

ingwinlu:
Just to clarify the relation to earlier work: this is not a new duplicate — it's the same branch, rebased onto current main after you merged the shared-subscription storage subset as #1564. The two commits that overlapped with #1564 were dropped during the rebase, so what remains here is only the drain-on-resume redelivery work (plus the FIFO/64-bit-pipe-id qos_db fixes it depends on). Nothing in this PR is already on main.

ingwinlu:
**Status after rebase (reopened):** this is the *redelivery/drain* half of the offline shared-sub QoS work. The *storage* half merged as #1564 and the qos_db leak fix as #1573, so I've rebased this branch onto current `main` — those overlapping commits dropped out, leaving only the drain: `NEW drain persisted QoS backlog on session resume` plus supporting qos_db fixes.

The one shared spot is the resend-getopt loop, which #1573 also touched — post-rebase they compose cleanly (pointer-tag strip → #1573's expired-msg free → new `req->drain` path). No conflicts.

I closed this yesterday only to land the prerequisites first; reopening now that they're in. Note thread-sanitize will stay red until the http teardown fix lands (I closed #1574 since you'd rather do that in `conf_fini`) — happy to hold this behind your fix or rebase again once it's in.

JaylinYu:
frankly speaking, nothing glaringly wrong in this PR.
But... it is not the right way for this. This is a common issue of AI generated PR, it doesn't understand the whole picture and the essence of the architecture, making it more and more complicated and hard to maintain.
I will implement it for you, a faster QoS resend, controlled by client' ACK reply.

ingwinlu:
I can see that adding the whole additional logic on top just for redelivery alone is not elegant. Thank you for looking into it.

What I can offer though is the id fixes and the tests as seperate commits? That way you can base your solution on it.

JaylinYu (changes requested):
remove commit da9684ba87ca5890eb233b7d2a3ce6af5ba59050
I will merge the rest.
You can expect the new qos resend policy will be ready in the next release at next week.

JaylinYu:
nevermind, I remove the last commit and merged

PR #1561, opened 2026-07-07
Fix SQLite compatiblity with qos msg cache

JaylinYu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened publish-PID extraction by adding safer bounds checks when reading message contents.
  * Improved MQTT QoS resend selection across TCP, TLS, and WebSocket by iterating cached candidates, skipping expired entries, and retrying until an eligible message is found.
  * Fixed WebSocket publish segment construction to prevent queued-length carryover between subscriptions.
  * Updated config size parsing to accept plain numeric values (bytes) without unit suffixes.
  * Improved clarity of MQTT-related logging output and switched SQLite QoS DB bindings to use 64-bit integers for packet/pipe IDs.

### DIFF
--- a/src/core/message.c
+++ b/src/core/message.c
@@ -793,16 +793,23 @@ uint16_t
 nni_msg_get_pub_pid(nni_msg *m)
 {
 	uint16_t pid;
-	uint8_t *pos, len;
+	uint16_t len;
+	uint8_t *pos;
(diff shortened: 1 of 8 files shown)


JaylinYu:
@coderabbitai review

JaylinYu:
@coderabbitai review again

PR #1564, opened 2026-07-14
* FIX [broker] store offline shared-subscription QoS msgs for cached sessions

ingwinlu:
## Summary

QoS-1/2 messages published while a shared-subscription (`$share/<group>/<filter>`) persistent session (`clean_start=false`) is offline are dropped instead of stored: the transport offline-cache branch re-matches the publish topic against the raw stored filter, so the `$share` prefix never matches and the message is freed. A resumed session gets none of its offline messages.

No group-level offline queue is involved — the dispatcher has already round-robined the message to exactly one (possibly offline) group member and handed it to that member's pipe, matching the per-Client-ID storage model. Only the final topic-match in the cache branch fails.

## Fix

A `shared_filter_skip()` helper strips `$share/<group>/` before `topic_filtern` in the offline-cache branch of the tcp/tls/ws transports — the same strip every online send path already does. Malformed shapes (`$share/`, `$share/g`, `$share/g/`) are returned unchanged.

A second commit hardens the same loop: it resolved the stored QoS from the *first* matching subscription and broke out. With shared filters now participating in the match, a session holding an overlapping pair (e.g. `$share/g/t` at QoS 0 subscribed before `t` at QoS 1) could resolve a QoS 1 publish to effective QoS 0 and drop it instead of storing. The loop now accumulates the maximum effective QoS over all matching subscriptions (single-copy delivery at the strongest match, per MQTT-3.3.4-2).

This is the minimal storage fix split out of #1560 as discussed in nanomq/nanomq#2355 ("your PR is welcome"). Redelivery pace is untouched — resumed sessions drain on the existing resend timer; the opt-in resume-pacing change will come separately.

## Validation

- Unit test for the helper covering plain filters, `$SYS` topics, well-formed and malformed `$share` shapes (`mqtt_parser_test`).
- The companion nanomq PR (nanomq/nanomq#2378) adds an end-to-end regression: {plain-control, `$share`, overlap, QoS-0 control} × {SQLite, in-memory}; after this fix shared members behave identically to plain subscribers on resume, and an overlapping `$share` QoS-0 + plain QoS-1 pair stores at QoS 1.

Related: nanomq/nanomq#2355

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MQTT shared subscription handling across TCP, TLS, and WebSocket transports.
  * Shared subscription prefixes are now correctly ignored when matching published topics and determining delivery settings.
  * Added handling for malformed, non-shared, wildcard, and empty subscription filters.

* **Tests**
  * Added coverage for shared subscription filter parsing and topic matching.

### DIFF
--- a/include/nng/protocol/mqtt/mqtt_parser.h
+++ b/include/nng/protocol/mqtt/mqtt_parser.h
@@ -140,6 +140,7 @@ NNG_DECL int  nmq_subinfo_decode(nng_msg *msg, void *l, uint8_t ver);
 NNG_DECL int  nmq_unsubinfo_decode(nng_msg *msg, void *l, uint8_t ver);
 NNG_DECL bool topic_filter(const char *origin, const char *input);
 NNG_DECL bool topic_filtern(const char *origin, const char *input, size_t n);
+NNG_DECL const char *shared_filter_skip(const char *filter);
 
 NNG_DECL int nmq_auth_http_connect(conn_param *cparam, conf_auth_http *conf);
 
--- a/src/sp/protocol/mqtt/mqtt_parser.c
+++ b/src/sp/protocol/mqtt/mqtt_parser.c
@@ -2116,6 +2116,31 @@ topic_filtern(const char *origin, const char *input, size_t n)
 }
 
 /**
+ * @brief skip the "$share/<group>/" prefix of a shared
+ * 		  subscription filter, so that the remainder can be
+ * 		  used for topic matching (topic_filter/topic_filtern)
+ *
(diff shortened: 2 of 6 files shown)


JaylinYu (approved):

JaylinYu:
I see. Fable5 is really strong. I have seen much PRs made by other AI. this one is clean and tidy.

ingwinlu:
it still gets lost if you let if run with too many options. but for something "small" like this its a time saver.

PR #1566, opened 2026-07-14
mqtt: limit logging on saturated broker queues

### LINKED ISSUE #1567: MQTT egress warnings can amplify queue saturation
**Describe the bug**

When a broker MQTT subscriber cannot drain its egress queue, the queue-full
path can log a warning for every dropped QoS 0 message while holding the pipe
lock. This can turn overload into synchronous log amplification.

**Expected behavior**

Queue saturation should remain observable without producing one synchronous
warning per dropped message or extending the pipe-lock critical section for
logging and message destruction.

**Actual behavior**

In a two-physical-core QoS 0 fan-in reproduction at 48k msg/s offered, the
latest NanoMQ source delivered about 2.5k msg/s at `warn` while emitting 528k
warnings. The same build at `error` delivered about 13k msg/s. Disabling only
the queue-full log raised receive rate to 17.6–18.5k msg/s.

**To Reproduce**

Use 32–48 emqtt-bench publishers, one emqtt-bench subscriber, a shared topic,
QoS 0, 256-byte payloads, `-I 1`, and a broker pinned to two physical cores.
Set `max_mqueue_len = 65535` and log level `warn`. Once the subscriber egress
queue saturates, count the queue-full warnings and compare receive rate with a
run at log level `error`.

**Environment Details**

- NanoNNG `main` embedded by the current NanoMQ source tree.
- Linux on an Intel i7-3770; broker pinned to CPUs 0,1.
- emqx/emqtt-bench MQTT v5 clients in high-frequency socket mode (`-A true`).

**Additional context**

The receive-side `no ctx found` warning has a similar burst pattern. A
disappeared destination pipe can also occur during a normal publication versus
disconnect race, where a warning may not be actionable.

yoch:
fixes #1567 MQTT egress warnings can amplify queue saturation

## Context

In a fan-in setup where one MQTT subscriber cannot drain its egress queue fast
enough, the current path appears to emit one warning for every dropped QoS 0
message. The warning and the message free also appear to happen while the pipe
lock is held.

On a two-core, 48k msg/s fan-in reproduction, this correlated with about 2.5k
msg/s received at `warn` and 528k warnings, compared with roughly 13k msg/s at
`error`. This suggests that synchronous warning output may amplify normal
overload into a much larger throughput cliff.

The `no ctx found` warning seems to have a similar pattern: its global rotate
counter is racy and can produce long warning bursts. A destination pipe that
disappears during publication also looks like an expected drop race rather than
an operator warning.

## Proposed change

This patch would:

- keep warning output enabled by default, but limit queue-full output to one
  warning per affected connection;
- expose `NMQ_OPT_MQTT_LOG_DROPS` so an embedding application could disable
  drop warnings without disabling accounting;
- move queue-full logging and message destruction out of the pipe lock;
- limit the no-context warning once per socket;
- lower the disappeared-pipe message to debug level;
- expose read-only transport sent/drop counters. Sends would be counted only
  after successful PUBLISH transport completion; drops would include full
  queues, disappeared pipes without a QoS database, and resize failures.

The existing FIFO behavior and QoS 0 drop policy would remain unchanged.

## Validation

I first bisected the hot path with 48k msg/s offered:

| Variant | Receive rate |
| --- | ---: |
| stock, warning enabled | ~2.5k msg/s |
| queue-full log disabled only | 17.6–18.5k msg/s |
| log/free moved outside the lock only | 8.0–9.4k msg/s |
| both changes | 13.8–18.6k msg/s |

With this patch and a 65,535-message queue, the QoS 0 fan-in runs reached
15,998, 23,896, and 31,905–31,948 msg/s for 16k, 24k, and 32k offered. The 32k
`warn` and `error` results were effectively the same, and a warning run emitted
one saturation warning. A 1-to-32 fan-out control reached 31,999 msg/s.

As a configuration-path control at 96k offered, enabling/disabling
`log_dropped_messages` resulted in 15,172/14,980 msg/s respectively, while the
warning count changed from one to zero. This would suggest that disabling the
warning changes observability rather than the drop mechanism.

At loads above 32k, a remaining single-pipe drain limit is still visible and
results are more variable. This patch is therefore intended to make overload
observable and to avoid warning amplification; it would not attempt to change
backpressure or queueing policy.

Release and ASAN builds passed. Tests cover the socket options, read-only
counters, configuration parsing, and public send completion; the MQTT test was
also repeated 20 times under ASAN.

## Summary by CodeRabbit

- **New Features**
  - Added MQTT metrics for tracking sent and dropped messages.
  - Added an MQTT option to enable or disable dropped-message logging.
  - Added the `log_dropped_messages` configuration setting, enabled by default.
- **Bug Fixes**
  - Improved handling of messages dropped when delivery queues are unavailable or full.
  - Reduced repetitive warnings by limiting duplicate drop and connection-context logs.
- **Configuration**
  - Added support for configuring dropped-message logging through MQTT settings.

### DIFF
--- a/include/nng/protocol/mqtt/nmq_mqtt.h
+++ b/include/nng/protocol/mqtt/nmq_mqtt.h
@@ -27,6 +27,9 @@ NNG_DECL int nng_nmq_tcp0_open(nng_socket *);
 
 #define NMQ_OPT_MQTT_PIPES "mqtt-clients-pipes"
 #define NMQ_OPT_MQTT_QOS_DB "mqtt-clients-qos-db"
+#define NMQ_OPT_MQTT_MSGS_DROPPED "mqtt-messages-dropped"
+#define NMQ_OPT_MQTT_MSGS_SENT "mqtt-messages-sent"
+#define NMQ_OPT_MQTT_LOG_DROPS "mqtt-log-drops"
 
 #ifdef __cplusplus
 }
--- a/include/nng/supplemental/nanolib/conf.h
+++ b/include/nng/supplemental/nanolib/conf.h
@@ -648,6 +648,7 @@ struct conf {
 	char      *conf_file;		// config path
 	char      *url;
 	bool       enable;
+	bool       log_dropped_messages;
 	size_t     property_size;
 	size_t     msq_len;
     uint16_t   max_topic_alias;
--- a/src/sp/protocol/mqtt/CMakeLists.txt
+++ b/src/sp/protocol/mqtt/CMakeLists.txt
@@ -8,6 +8,7 @@ if (NNG_PROTO_MQTT_BROKER)
     nng_sources_if(NNG_PROTO_MQTT_BROKER mqtt_parser.c nmq_mqtt.c auth_http.c)
     nng_headers_if(NNG_PROTO_MQTT_BROKER nng/protocol/mqtt/mqtt_parser.h nng/protocol/mqtt/nmq_mqtt.h)
     nng_defines_if(NNG_PROTO_MQTT_BROKER NNG_HAVE_MQTT_BROKER)
+    nng_test(nmq_mqtt_test)
 endif ()
 
 nng_test(mqtt_parser_test)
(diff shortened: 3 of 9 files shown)


JaylinYu:
How about adding a simple rotate divider that prints one line of new log for every 100-1000  line of old logging?

@yoch

JaylinYu:
I believe there is a simpler way

PR #1575, opened 2026-07-16
* FIX [broker] resolve TSan data races behind the red sanitize job

ingwinlu:
## What

Fixes the deterministic ThreadSanitizer failures in `nng.mqtt_broker_tcp` and `nng.mqttv5_broker_tcp` that keep the `sanitize` CI job red on every commit. Three independent fixes:

1. **`nmq_mqtt`: make `nano_pipe->reason_code` atomic.** It is written from the recv/send/timer callbacks without `p->lk` and read in `nano_pipe_close` under it (the site previously carried a `dont mind the TSAN here` comment). Converted to `nni_atomic_int`, matching the neighbouring `closed` flag.

2. **`mqtt_parser`: don't re-read the refcount after `nni_atomic_dec_nv` in `conn_param_free`.** The `log_trace` argument performed a second atomic read after the decrement; a concurrent `conn_param_free` on another thread can drop the last reference and free the block in between, making the re-read a use-after-free. TSan catches this intermittently (`free` in `mqtt_tcptran_pipe_fini` on the reap thread vs. the re-read on the caller's thread). Reuse the value the decrement already returned.

3. **`tests/trantest.h`: receive broker messages synchronously.** The broker send/recv tests posted `nng_ctx_recv(work->ctx, work->aio)` and then read the message off the aio after `nng_msleep`. A sleep establishes no happens-before edge, so TSan reports the `nni_aio_get_msg` / `nni_msg_get_conn_param` / `nni_msg_get_pipe` reads as racing with the broker worker threads — and on a slow run the assertions can read a stale pointer. Switched to `nng_ctx_recvmsg()`, which waits on its own internal aio; `work->aio` remains send-only.

   Note for reviewers: waiting on `work->aio` directly is not possible because `nano_ctx_send` calls `nni_aio_begin` but never finishes the aio on any path (fire-and-forget). That latches the aio's task busy count, so any later `nng_aio_wait` on the same aio blocks forever even after a completed receive. Callback-driven apps never notice, but it diverges from the nng aio contract and rules out `nng_aio_wait`/`nng_ctx_sendmsg`-style usage with the broker protocol — possibly worth a follow-up discussion.

## What this does NOT fix

The third red test in the sanitize job, `nng.wss`, has a different root cause: `exit(EXIT_FAILURE)` in `nni_http_conn_fini` can run the `NNG_TEST_LIB` `atexit(nng_fini)` handler on the reap thread, tearing down `sp_tran_list` while the main thread is still using it (TSan: `nni_list_first` at `list.c:34`, followed by the `trantest.h:227 tt->tran != NULL` failure). That path is removed by #1574; once both land, the `thread` sanitizer matrix entry should be green.

## Verification

- Built with `CC=clang` per `sanitizer.yml` (`-DNNG_SANITIZER=thread -DCMAKE_BUILD_TYPE=Debug -DNNG_ENABLE_TLS=ON -DNNG_TOOLS=OFF`).
- `mqtt_broker_tcp` and `mqttv5_broker_tcp`: 0 TSan warnings and 0 test failures across 25+ runs each (previously 3-4 race reports per run, every run).
- Full `ctest` suite: failure set identical to unpatched `main` in the same environment (remaining local failures are environment artifacts plus the `wss` issue above).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety and correctness when handling MQTT pipe disconnect reason codes under concurrent conditions.
  * Enhanced diagnostic logging around reference-count cleanup during connection teardown to better reflect the latest state.

* **Tests**
  * Improved MQTT broker integration tests by switching to synchronous response collection, removing timing-based waits and reducing flakiness across CONNECT, SUBSCRIBE, PUBLISH, and UNSUBSCRIBE flows.

### DIFF
--- a/src/sp/protocol/mqtt/mqtt_parser.c
+++ b/src/sp/protocol/mqtt/mqtt_parser.c
@@ -922,7 +922,6 @@ conn_param_free(conn_param *cparam)
 		return;
 	}
 	if (nni_atomic_dec_nv(&cparam->refcnt) != 0) {
-		log_trace("%p is not freed yet!! %d", cparam, nni_atomic_get(&cparam->refcnt));
 		return;
 	}
 	log_trace("destroy conn param %p", cparam);
--- a/src/sp/protocol/mqtt/nmq_mqtt.c
+++ b/src/sp/protocol/mqtt/nmq_mqtt.c
@@ -77,7 +77,8 @@ struct nano_pipe {
 	nano_sock  *broker;
 	conn_param *conn_param;
 	nni_lmq     rlmq; // only for sending cache
-	uint8_t     reason_code;
+	// disconnect reason; set from cb threads, read while closing
+	nni_atomic_int reason_code;
 	uint32_t    id;  // pipe id of nni_pipe
 	uint16_t    rid; // index of packet ID for resending
 	uint16_t    keepalive;
@@ -222,7 +223,7 @@ nano_pipe_timer_cb(void *arg)
 				nni_id_remove(
 				    &s->cached_sessions, p->pipe->p_id);
 			}
-			p->reason_code = 0x8E;
+			nni_atomic_set(&p->reason_code, 0x8E);
 			nni_mtx_unlock(&sock->lk);
 			nni_pipe_close(p->pipe);
 			return;
@@ -247,7 +248,7 @@ nano_pipe_timer_cb(void *arg)
 		if (qos_backoff > 0) {
 			log_warn("Warning: close pipe %u & kick client due to "
 			         "KeepAlive timeout!", p->id);
-			p->reason_code = NMQ_KEEP_ALIVE_TIMEOUT;
+			nni_atomic_set(&p->reason_code, NMQ_KEEP_ALIVE_TIMEOUT);
 			nni_mtx_unlock(&p->lk);
 			nni_pipe_close(p->pipe);
 			return;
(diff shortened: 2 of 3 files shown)


src/sp/protocol/mqtt/mqtt_parser.c
```diff
@@ -921,8 +921,11 @@ conn_param_free(conn_param *cparam)
 	if (cparam == NULL) {
 		return;
 	}
-	if (nni_atomic_dec_nv(&cparam->refcnt) != 0) {
-		log_trace("%p is not freed yet!! %d", cparam, nni_atomic_get(&cparam->refcnt));
+	int refcnt = nni_atomic_dec_nv(&cparam->refcnt);
+	if (refcnt != 0) {
+		// do not touch cparam again: another thread may
+		// drop the last reference and free it concurrently
+		log_trace("%p is not freed yet!! %d", cparam, refcnt);
```
JaylinYu: just delete it. the log was added for debugging , and I forgot to remove it after fix the memleak.

JaylinYu (changes requested):

src/sp/protocol/mqtt/mqtt_parser.c
ingwinlu: Removed the log.

JaylinYu (approved):
thanks

PR #1591, opened 2026-07-29
FIX [mqtt_parser] own MQTT v5 topic aliases per connection

jayenashar:
## Problem

MQTT v5 topic aliases lived in a global table keyed on pipe id. nanomq
overwrites the pipe id with `nanomq_siphash_32(clientid)` in
`tcptran_pipe_init` (`src/sp/transport/mqtt/broker_tcp.c`), so that table is
really keyed per client id and is shared by successive connections of one
client.

The broker deleted the key when it processed the disconnect event. But the
transport PUBACKs a QoS 1 publish as soon as it reads it off the wire, before
any app layer processing, so a client can disconnect while publishes the
broker has already acknowledged are still queued on worker ctxs. When the
disconnect event won that race, those publishes failed their alias lookup,
were dropped despite having been acknowledged, and the pipe was kicked:

```
ERROR nanomq/pub_handler.c: could not find topic by alias: 10
WARN  src/sp/protocol/mqtt/nmq_mqtt.c: pipe id <publisher> is gone, pub failed
```

## Why not the CONNACK reset

nanomq/nanomq#2391 moved the reset to CONNACK. @JaylinYu closed it, and he was
right: because the key is a client id hash, an alias record for a client that
never reconnects is never reclaimed. He asked for the state to be bound to the
connection, or for reclamation to be deferred until queued publishes have
drained. This takes the first route.

## Fix

`conn_param` is refcounted, and every message of a connection still in flight
holds a reference: `nano_pipe_recv_cb` clones it for each `CMD_PUBLISH` before
handing the message to the app layer. Owning the alias vector there gives
exactly the lifetime we want, without any new bookkeeping:

- a publish the broker has already acknowledged cannot lose its mapping to a
  concurrent disconnect, because its own reference keeps the vector alive
- the vector is freed with the last reference, so nothing is left unattended
  when a client never connects back
- a new connection parses a new `conn_param`, so it always starts empty

The global khash and the deletion on the disconnect event are both gone.
`dbhash_{init,destroy}_alias_table`, `dbhash_insert_atpair`,
`dbhash_find_atpair` and `dbhash_del_atpair_queue` are replaced by
`dbhash_atpair_insert/find/free_all`, which operate on a caller owned vector.
`conn_param_set_topic_alias` and `conn_param_get_topic_alias` wrap them with a
per `conn_param` mutex, since several worker ctxs can be handling publishes of
one connection at once, and they copy the topic out under that lock, which
`dbhash_find_atpair` did not do.

`nni_get_conn_param_from_msg` builds a `conn_param` field by field and does not
call `conn_param_init`, so it initialises the new members too.

### What this costs

Aliases are no longer shared between connections, which is the point, but it is also a memory trade worth stating. Previously every connection whose client id hashed to the same pipe id shared one vector; now each connection owns its own, so worst case alias memory is the number of live connections times up to `max_topic_alias` topics, each bounded by the packet size limit. Any single client could already pin that much for the life of its connection, so the ceiling per connection is unchanged, and the vector is freed with the connection rather than lingering under a client id key. What goes away is the accidental sharing, which was never correct.

## A second bug this fixes

MQTT 5 makes topic aliases per Network Connection state (spec 3.3.2.3.4,
MQTT-3.3.2-7). Today they survive a reconnect: the key is a client id hash, and
the cached session path sets `p->event = false` so no disconnect event is
generated to clear them. A client reconnecting with `clean_start=0` therefore
inherits the previous connection's aliases.

Register alias 9 for `per/a`, disconnect cleanly, reconnect with the same
client id and `clean_start=0` plus session expiry, then publish with an empty
topic name and alias 9:

- current `main`: the message is delivered on `per/a`
- this branch: the alias is rejected with `could not find topic by alias: 9`

## Validation

Local build on top of `8a33434a`, broker pinned to 2 cores, raw socket MQTT v5 clients.

The disconnect race does not reproduce naturally at this scale, so to make it observable I added an identical env gated `nng_msleep()` immediately before the alias lookup in `handle_pub` on both the unpatched and patched build. It stands in for a worker that has been handed a publish but has not run yet. With a 50 ms delay, 4 workers, burst of 10, 10 trials each:

- unpatched: 8 `could not find topic by alias`, 345 of 440 publishes delivered
- patched: 0, 440 of 440

On the shipped code without the delay hook, under an AddressSanitizer build (`-DDEBUG=1 -DASAN=1`, both the broker and the nng library instrumented): 2520 of 2520 aliased publishes delivered, 660 of 660 through a same client id takeover, the only two `could not find topic by alias` lines in the whole broker output being the two a stale alias probe deliberately provokes, and no ASan reports.

Under a ThreadSanitizer build of the same code, which is the interesting one since this change adds a mutex: 960 aliased publishes across 6 concurrent workers with hard disconnects, plus same client id takeovers and a kept session reconnect, produced **no data races at all**, and none naming `topic_alias`, `dbhash_atpair` or the accessors. The TSan runtime was confirmed live via its own startup banner rather than inferred from an empty report.

Per connection semantics hold, both for a clean reconnect and for a kept session. Memory is flat against `main`: 200 connections registering 200 aliases each with 200 byte topics ends at 137980 kB RSS patched versus 137528 kB unpatched, where an unfreed alias set would have been several MB.

`hash_test` passes, and bridging between two brokers still forwards, which
exercises the `nni_get_conn_param_from_msg` path.

## Residual

This fixes the lifetime, not the ordering. App layer publish processing is still not strictly ordered per connection, so an extreme overload reorder of the alias registering publish against a later lookup on the same connection remains theoretically possible. I did not observe it in 400 isolation probe trials, and it is a separate problem from the one here: closing it would need per connection ordering of `handle_pub`, which is the shape of the "defer reclamation until queued publishes have drained" option and a much larger change. Flagging it so the scope of this PR is not overstated.

## Companion

The nanomq side (`broker.c`, `pub_handler.c`, the submodule bump and a
regression test for the alias lookup path, which nothing covers today) is ready
on `jayenashar/nanomq:fix/topic-alias-conn-param` and I will raise it once this
lands.

## Summary by CodeRabbit

* **New Features**
  * Added MQTT topic alias support for connections.
  * Applications can register topic aliases and retrieve their associated topics.
  * Topic aliases are maintained independently for each connection.
  * Re-registering an alias updates its associated topic.
* **Bug Fixes**
  * Improved alias handling for concurrent MQTT message processing.
* **Tests**
  * Expanded coverage for alias registration, lookup, replacement, empty states, and cleanup.

### DIFF
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -1684,7 +1684,8 @@ NNG_DECL void           conn_param_set_password(
 NNG_DECL void        conn_param_set_proto_ver(conn_param *cparam, uint8_t ver);
 // Topic aliases belong to the connection that registered them (MQTT 5 spec
 // 3.3.2.3.4). conn_param_get_topic_alias returns a copy owned by the caller,
-// or NULL if this connection never registered that alias.
+// which releases it with nng_strfree(), or NULL if this connection never
+// registered that alias.
 NNG_DECL void        conn_param_set_topic_alias(
            conn_param *cparam, uint32_t alias, const char *topic);
 NNG_DECL char       *conn_param_get_topic_alias(
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -1682,6 +1682,13 @@ NNG_DECL void           conn_param_set_username(
 NNG_DECL void           conn_param_set_password(
               conn_param *cparam, const char *password);
 NNG_DECL void        conn_param_set_proto_ver(conn_param *cparam, uint8_t ver);
+// Topic aliases belong to the connection that registered them (MQTT 5 spec
+// 3.3.2.3.4). conn_param_get_topic_alias returns a copy owned by the caller,
+// or NULL if this connection never registered that alias.
+NNG_DECL void        conn_param_set_topic_alias(
+           conn_param *cparam, uint32_t alias, const char *topic);
+NNG_DECL char       *conn_param_get_topic_alias(
+          conn_param *cparam, uint32_t alias);
 NNG_DECL uint64_t    conn_param_get_will_delay_timestamp(conn_param *cparam);
 NNG_DECL uint64_t    conn_param_get_will_mexp(conn_param *cparam);
 NNG_DECL void        nng_msg_set_proto_data(nng_msg *m, void *ops, void *data);
--- a/include/nng/supplemental/nanolib/hash_table.h
+++ b/include/nng/supplemental/nanolib/hash_table.h
@@ -52,17 +52,20 @@ alias_cmp(void *x_, void *y_)
 	return *alias - ele_x->alias;
 }
 
-NNG_DECL void dbhash_init_alias_table(void);
+// Topic aliases are per-connection state (MQTT 5 spec 3.3.2.3.4), so the
+// vector of alias/topic pairs belongs to the connection that registered them
+// (its conn_param) rather than to a global table keyed on pipe id. A NULL
+// vector is an empty one, and the owner is responsible for serialising
+// access; see conn_param_set_topic_alias().
 
-NNG_DECL void dbhash_destroy_alias_table(void);
 // This function do not verify value of alias and topic,
 // therefore you should make sure alias and topic is
 // not illegal.
-NNG_DECL void dbhash_insert_atpair(uint32_t pipe_id, uint32_t alias, const char *topic);
+NNG_DECL void dbhash_atpair_insert(dbhash_atpair_t ***vecp, uint32_t alias, const char *topic);
 
-NNG_DECL const char *dbhash_find_atpair(uint32_t pipe_id, uint32_t alias);
+NNG_DECL const char *dbhash_atpair_find(dbhash_atpair_t **vec, uint32_t alias);
 
-NNG_DECL void dbhash_del_atpair_queue(uint32_t pipe_id);
+NNG_DECL void dbhash_atpair_free_all(dbhash_atpair_t **vec);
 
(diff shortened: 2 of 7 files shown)


jayenashar:
Thanks, took the first one and skipped the second.

**Deallocator doc**: applied. The declaration now names `nng_strfree()`, which matches the implementation. Worth having on a new public API, since reaching for plain `free()` on an `nni_alloc` backed string would be a genuine mismatch.

**Shared init helper**: skipping, though the drift you point at is real rather than hypothetical. The second commit in this PR exists precisely because I initialised the new members in `conn_param_init` and missed `nni_get_conn_param_from_msg`, which would have left `conn_param_free` finalising an uninitialised mutex.

The reason I am not taking it is that a helper for these two lines treats the symptom. `nni_get_conn_param_from_msg` already hand duplicates around twenty initialisations from `conn_param_init` (`refcnt`, the TLS fields, `properties`, `will_properties`, every `mqtt_string`), so factoring out only the topic alias pair singles out the newest two fields and leaves every other field exposed to exactly the same drift. It also needs a new helper in a shared header, since the two constructors live in different translation units.

The fix that would actually close the gap is having `nni_get_conn_param_from_msg` call `conn_param_init` and then override what it needs, which is a wider change than this PR should carry and is better raised on its own.

JaylinYu:
This is a big change, requires my a lot attention to review and test, please allow me processing this later

jayenashar:
no rush, just trying to Make Actions Green Again

PR #1592, opened 2026-07-30
mqtt: reject unsuccessful CONNACK negotiation

kazazes:
## Summary

- reject nonzero MQTT CONNACK reason codes during client transport negotiation
- close and reap the rejected connection with `NNG_ECONNREFUSED`
- apply the behavior consistently to TCP, TLS, and QUIC transports
- add a local TCP regression test that verifies the refusal and preserves the MQTT reason code

## Root cause

The MQTT transport negotiation callback currently uses the decoder result as
the negotiation result. A syntactically valid CONNACK therefore completes
negotiation successfully even when its MQTT return/reason code rejects the
CONNECT request.

That publishes a rejected connection as an active pipe. If the broker does not
close the network connection promptly, the dialer never enters its retry path
and sends can remain stuck behind a pipe that was never accepted at the MQTT
layer.

MQTT defines only reason code 0 as a successful CONNACK. Returning
`NNG_ECONNREFUSED` for every other code keeps the original MQTT reason available
through `NNG_OPT_MQTT_CONNECT_REASON` while letting the existing dialer
backoff/reconnect behavior handle the failed attempt.

This was observed with MQTT 5 reason code `0x87` (`Not authorized`): after the
credential was accepted later, the client remained on the previously rejected
pipe instead of reconnecting.

## Validation

- `cmake -S . -B build-pr36 -G Ninja -DNNG_TESTS=ON -DNNG_ENABLE_TLS=OFF -DNNG_ENABLE_QUIC=OFF -DCMAKE_BUILD_TYPE=Debug`
- full default build completed
- `ctest --test-dir build-pr36 -R mqtt_tcp_test --output-on-failure`
- `cmake -S . -B build-pr36-quic -G Ninja -DNNG_TESTS=OFF -DNNG_ENABLE_TLS=OFF -DNNG_ENABLE_QUIC=ON -DCMAKE_BUILD_TYPE=Debug`
- full QUIC-enabled build completed, including `mqtt_quic.c`
- full `ctest --output-on-failure`: 103/104 tests passed; the sole failure, `nng.supplemental.conf_test`, reproduces unchanged on `upstream/main`

The regression test runs a local raw TCP listener, returns a rejected MQTT
3.1.1 CONNACK without relying on the peer to close first, and verifies that the
dial fails with `NNG_ECONNREFUSED` while the original CONNACK code remains
queryable.

## Summary by CodeRabbit

- **Bug Fixes**
  - MQTT connections now fail promptly when a broker rejects the CONNECT request.
  - TCP, TLS, and QUIC transports report connection refusal while preserving the broker’s specific MQTT reason code.
  - Improved cleanup prevents rejected connections from proceeding as if negotiation succeeded.

- **Tests**
  - Added coverage confirming unauthorized CONNACK responses fail negotiation and expose the correct reason code.


JaylinYu (approved):
THanks

PR #1600, opened 2026-08-03
FIX [http_conn] make connection error cleanup safe

### LINKED ISSUE #1599: HTTP connection error cleanup is unsafe
### Environment

- Platform: macOS
- Branch: `main` (`f2acbcf251f357a13752b0de42243cd90912bade`)
- CMake configuration: `-DNNG_TESTS=ON -DNNG_ENABLE_TLS=OFF`

### Observed behavior

`http_init()` uses the runtime `nni_http_conn_fini()` path when buffer or AIO allocation fails, even though the connection is only partially initialized. The finalizer unconditionally schedules the read and write AIOs, so a null AIO can be dereferenced.

If scheduling the front-end cleanup AIO fails, `http_init()` calls `exit(EXIT_FAILURE)` instead of returning the error to the caller.

During runtime teardown, a stopped read or write AIO cannot install its cancellation callback. The current code logs that scheduling failure but does not mark that direction closed, so the front-end AIO may never be reaped after the other callback runs.

### Expected behavior

Initialization failures should synchronously free only the state that was successfully initialized and return an error. Runtime teardown should account for failed read/write cancellation scheduling and reap the front-end AIO exactly once when both directions are closed. Library code should not terminate the process for an AIO scheduling error.

### Reproduction

The front-end scheduling failure can be exercised deterministically by setting the freshly allocated cleanup AIO timeout to zero before scheduling it. Partial allocation and stopped-AIO paths follow the same cleanup code described above.

bmiesch:
## Summary
- clean up partially initialized HTTP connections synchronously
- return front-end cleanup AIO scheduling errors instead of terminating the process
- account for stopped read/write AIOs so final connection cleanup runs exactly once
- add regression coverage for initialization and runtime teardown failures

## Context
HTTP initialization currently delegates failures to the runtime finalizer, which assumes all AIOs exist. Runtime teardown also leaves the front-end cleanup AIO scheduled when a stopped read or write AIO cannot install its cancellation callback.

## Testing
- `cmake -S . -B build-http-cleanup -G Ninja -DNNG_TESTS=ON -DNNG_ENABLE_TLS=OFF -DCMAKE_BUILD_TYPE=Debug`
- `cmake --build build-http-cleanup -j4`
- `ctest --test-dir build-http-cleanup --output-on-failure -j4` (103/105 passed; two fixed-port address conflicts)
- `ctest --test-dir build-http-cleanup -R "^(nng.multistress|nng.mqttv5_broker_tcp)$" --output-on-failure -j1` (2/2 passed)
- focused `nng.supplemental.http_conn_test` regression test passes

## Issues
Closes #1599

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP connection cleanup when initialization or asynchronous operations fail.
  * Prevented potential resource leaks and ensured connections finalize correctly when writes are stopped.
  * Preserved timeout errors instead of terminating unexpectedly.

* **Tests**
  * Added coverage for immediate timeout failures and stopped-write cleanup scenarios.
  * Verified that connection finalization occurs exactly once.

### DIFF
--- a/src/sp/protocol/mqtt/auth_http.c
+++ b/src/sp/protocol/mqtt/auth_http.c
@@ -257,8 +257,8 @@ send_request(conf_auth_http *conf, conf_auth_http_req *conf_req,
 	// The Host: header is already set up too.
 	set_data(req, conf_req, params);
 	// Send the request, and wait for that to finish.
-	nng_http_conn_write_req(conn, req, aio);
 	nng_aio_set_timeout(aio, conf->timeout * 1000);
+	nng_http_conn_write_req(conn, req, aio);
 	nng_aio_wait(aio);
 
 	if ((rv = nng_aio_result(aio)) != 0) {
--- a/src/supplemental/http/http_conn.c
+++ b/src/supplemental/http/http_conn.c
@@ -2,7 +2,7 @@
 // Copyright 2021 Staysail Systems, Inc. <info@staysail.tech>
 // Copyright 2018 Capitar IT Group BV <info@capitar.com>
 // Copyright 2019 Devolutions <info@devolutions.net>
-// Copyright 2026 Brendan Miesch <brendan@picogrid.com>
+// Copyright 2026 NanoMQ Team, Inc.
 //
 // This software is supplied under the terms of the MIT License, a
 // copy of which should be located in the distribution where this
--- a/src/supplemental/http/http_conn.c
+++ b/src/supplemental/http/http_conn.c
@@ -2,6 +2,7 @@
 // Copyright 2021 Staysail Systems, Inc. <info@staysail.tech>
 // Copyright 2018 Capitar IT Group BV <info@capitar.com>
 // Copyright 2019 Devolutions <info@devolutions.net>
+// Copyright 2026 Brendan Miesch <brendan@picogrid.com>
 //
 // This software is supplied under the terms of the MIT License, a
 // copy of which should be located in the distribution where this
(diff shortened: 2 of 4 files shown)


src/sp/protocol/mqtt/auth_http.c
```diff
@@ -257,8 +257,8 @@ send_request(conf_auth_http *conf, conf_auth_http_req *conf_req,
 	// The Host: header is already set up too.
 	set_data(req, conf_req, params);
 	// Send the request, and wait for that to finish.
-	nng_http_conn_write_req(conn, req, aio);
 	nng_aio_set_timeout(aio, conf->timeout * 1000);
+	nng_http_conn_write_req(conn, req, aio);
```
JaylinYu, line 261: whoops

src/supplemental/http/http_conn.c
```diff
@@ -2,6 +2,7 @@
 // Copyright 2021 Staysail Systems, Inc. <info@staysail.tech>
 // Copyright 2018 Capitar IT Group BV <info@capitar.com>
 // Copyright 2019 Devolutions <info@devolutions.net>
+// Copyright 2026 NanoMQ Team, Inc.
```
JaylinYu, line 5: thanks

src/supplemental/http/http_conn.c
```diff
@@ -750,20 +808,25 @@ http_init(nni_http_conn **connp, nng_stream *data)
 	nni_aio_list_init(&conn->wrq);
 
 	if ((conn->rd_buf = nni_alloc(HTTP_BUFSIZE)) == NULL) {
-		nni_http_conn_fini(conn);
+		http_init_fini(conn);
```
JaylinYu, line 811: I dont see this is necessary. What's the problem with old fini func?

src/supplemental/http/http_conn.c
```diff
@@ -712,30 +747,53 @@ nni_http_conn_setopt(nni_http_conn *conn, const char *name, const void *buf,
...
+		nni_aio_free(wr_aio);
+		nni_aio_free(rd_aio);
 	} else {
-		nni_aio_reap(conn->wr_aio);
-		nni_aio_reap(conn->rd_aio);
+		nni_aio_reap(wr_aio);
+		nni_aio_reap(rd_aio);
+	}
+	if (fe_aio != NULL) {
+		nni_aio_reap(fe_aio);
```
JaylinYu, line 782: why so rush? we can just rely on wr_aio/rd_aio cancel cb

JaylinYu (commented):
need more time to test & review

src/supplemental/http/http_conn.c
JaylinYu: Ok, I see the problem. I suggest to keep both

JaylinYu (approved):

JaylinYu:
what AI model are you using for this PR? Pretty curious.

PR #1602, opened 2026-08-04
fix: add strict Base64 boundary decoder

reneleonhardt:
## Summary

- Add the `nmq_base64_*` API and a strict padded Base64 decoder for NanoMQ consumers.
- Preserve the legacy `base64_*` API while rejecting invalid characters, misplaced padding, non-canonical pad bits, and insufficient output space.
- Make the decoder reject high-bit input safely and make its size macros evaluate caller expressions only once.
- Add focused boundary, size-calculation, and out-of-tree configuration-test coverage; provide the test fixtures through CMake and avoid null dereferences when a fixture cannot be loaded.
- Close the explicit TLS listener in the port-zero bind test and correct its address assertion so sanitizer runs are leak-free.
- Keep Darwin coverage on Homebrew `mbedtls@3` and restrict the coverage workflow permissions to the required read-only scope.

## Validation

The full macOS CTest suite passes with `NNG_TESTS=ON`: 108/108 tests passed, including the legacy and strict Base64 tests, `conf_test`, and the TLS transport tests. The address-sanitized TLS transport test also passes.

## New Features
* Added Base64 encoding and decoding support, including strict validation for padded input.
* Added overflow-safe output-size calculations and clearer handling of invalid input, malformed padding, and insufficient output buffers.
* Improved compatibility for Base64 APIs when used from C++.

## Bug Fixes
* Improved Base64 decoding reliability for bytes with the high bit set.
* Updated macOS coverage builds to use a compatible mbedTLS 3.x release.

## Tests
* Added coverage for valid and invalid Base64 data, buffer limits, size-calculation edge cases, and configuration path handling.


JaylinYu:
we are going to upgrade the mbedtls dependency
https://github.com/nanomq/nanomq/issues/2319

reneleonhardt:
Great to hear, all those outdated dependencies are driving me crazy 😅

include/nng/supplemental/nanolib/nmq_base64.h
```diff
@@ -1,30 +1,71 @@
 #ifndef NMQ_BASE64_H
 #define NMQ_BASE64_H
+
 #include "nng/nng.h"
 
-#define BASE64_ENCODE_OUT_SIZE(s)                                           \
```
JaylinYu, line 5: I would rather stay this way for better performance.

JaylinYu (approved):

include/nng/supplemental/nanolib/nmq_base64.h
JaylinYu: But I guess your usecase is special and need it

reneleonhardt: Do I have to change anything now after merge?

You can always suggest changes in the GitHub UI.

I don't know why "Allow edits and access to secrets by maintainers" wasn't checked, I never change the default, usually it's allowed.

PR #1604, opened 2026-08-05
Update query parquet  by exchange name instead of topic

alvin1221:
1. Update query parquet  by exchange name instead of topic;
2. Support replacing ${VIN} in the exchange topic.


JaylinYu (approved):
not much work

PR #1608, opened 2026-08-09
Restore conn_param ref counter method

JaylinYu:
We have to decouple transport & protocol on the shared conn_param. I used to make it right, but accidentally break it at last ver.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection lifecycle handling across MQTT, secure MQTT, and WebSocket transports.
  * Prevented connection settings from being released prematurely or redundantly during connection shutdown.
  * Increased reliability when closing and reusing transport connections.


JaylinYu:
@coderabbitai review

PR #1620, opened 2026-08-24
FIX [tls] flush final OpenSSL handshake flight

### LINKED ISSUE #1619: OpenSSL TLS server can complete before flushing the final handshake flight
**Describe the bug**

The OpenSSL TLS engine can report a successful server handshake while encrypted handshake bytes remain pending in the connection write BIO. `open_conn_handshake()` jumps to its `finished` path immediately after `SSL_do_handshake()` makes `SSL_is_init_finished()` true, before `ec->wbio` is drained.

**Expected behavior**

All pending write-BIO data is sent to the peer before the engine reports handshake completion.

**Actual behavior**

With mutual TLS, the client sends its certificate and Finished message, then waits until timeout because the server's final handshake flight is never sent. The server reports the handshake as successful.

**To Reproduce**

1. Build with:

   ```text
   -DNNG_ENABLE_TLS=ON
   -DNNG_TLS_ENGINE=open
   ```

2. Start a TLS listener configured with a server certificate, a CA chain for the client certificate, and `NNG_TLS_AUTH_MODE_REQUIRED`.
3. Connect with a TLS client presenting the trusted client certificate.
4. Observe that the client writes Finished but never reads the server Finished message.

The same behavior is visible with an OpenSSL client:

```text
SSL_connect:SSLv3/TLS write finished
<timeout waiting for server final flight>
```

Moving the completion check until after the write BIO is drained makes the handshake complete normally.

**Environment Details**

- NanoNNG: current `master` and commit `ebd4b8424735f72652b625f9167663683164aafd`
- Operating system: Linux ARM64
- Compiler: GCC 12
- OpenSSL: 3.6
- Testing scenario: TLS 1.2 mutual authentication with a required client certificate

**Client SDK**

Reproduced with `openssl s_client` and with NanoNNG TLS stream clients.

**Additional context**

A focused mutual-TLS regression test can use the existing `nuts_server_crt`, `nuts_server_key`, `nuts_client_crt`, and `nuts_client_key` fixtures.

bmiesch:
fixes #1619 OpenSSL TLS server can complete before flushing the final handshake flight

`open_conn_handshake()` could jump to its success path as soon as `SSL_is_init_finished()` became true while encrypted handshake bytes were still pending in the write BIO. A mutual-TLS client would send its Finished message and then wait until timeout for the server's final flight.

Move the completion check after the write BIO drain so all pending handshake data is sent before success is reported.

Validation:

- configured and built current `main` with `-DNNG_ENABLE_TLS=ON -DNNG_TLS_ENGINE=open -DNNG_TESTS=ON`
- built the `tls_test` target
- passed `tls conn refused`
- reproduced a TLS 1.2 mutual-authentication timeout before the change
- verified the mutual-authentication handshake completes after the change with certificate verification successful

By submitting this pull request, I confirm that I have read and agree to the repository's contributing guidelines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS connection-handshake handling to ensure completion is confirmed after required network processing.
  * Improved reliability for secure connections during handshake operations.
  * TLS protocol version settings are now correctly applied and invalid minimum versions are rejected.

* **Tests**
  * Added coverage for mutual TLS authentication using client and server certificates.

### DIFF
--- a/src/supplemental/tls/openssl/openssl.c
+++ b/src/supplemental/tls/openssl/openssl.c
@@ -269,7 +269,10 @@ open_conn_close(nng_tls_engine_conn *ec)
 static int
 open_conn_handshake(nng_tls_engine_conn *ec)
 {
+	int ensz;
 	int rv;
+	int sz;
+
 	print_trace();
 	if (ec->ok == 1)
 		return 0;
@@ -281,7 +284,6 @@ open_conn_handshake(nng_tls_engine_conn *ec)
 				"openssl handshake still in process rv%d", rv);
 	}
 	if (rv == SSL_ERROR_WANT_READ || rv == SSL_ERROR_WANT_WRITE) {
-		int ensz, sz;
 		while ((ensz = open_net_read(ec->tls, ec->wbuf, OPEN_BUF_SZ)) > 0) {
 			sz = BIO_write(ec->rbio, ec->wbuf, ensz);
 			log_debug("NNG-TLS-CONN-HANDSHAKE" "BIO write sz%d/%d", sz, ensz);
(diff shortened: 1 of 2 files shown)


bmiesch:
The failing address-sanitizer job is in the existing `nng.tcp` transport test and reports allocations from `src/sp/transport/tcp/tcp.c`; this PR only changes `src/supplemental/tls/openssl/openssl.c`.

The sanitizer workflow installs mbedTLS and configures `-DNNG_ENABLE_TLS=ON` without `-DNNG_TLS_ENGINE=open`, so the changed OpenSSL engine is not compiled into that job. The three normal build jobs passed. This appears unrelated and should be resolved by rerunning the address-sanitizer job.

JaylinYu:
@coderabbitai review

JaylinYu:
not an easy issue to find

JaylinYu (approved):

PR #1621, opened 2026-08-24
Fixed null-deref when renaming file with MD5.

xinyi-xs:


JaylinYu (approved):

JaylinYu:
@xinyi-xs  remember using `git commit -s`

PR #1624, opened 2026-08-27
Fix will property propagation to publish

thebolt:
## Summary

MQTT v5 allows a different set of properties on the will message sent as part of the CONNECT message from what is allowed on PUBLISH or on the CONNECT message itself. This results in that some will messages, such as those including will_delay_interval will be accepted in CONNECT, but then result in an malformed message being published to clients, which in turn makes them close the connection.

The dropped connection has been verified both with mosquitto and paho MQTT-clients.

This change verifies that the will message does not contain properties not allowed by the standard, and makes sure the will_delay_interval specifically is not included in the set of properties forwarded to the PUBLISH path.

## Validation

The full test suite fails on two tests (nng.tcp and nng.ws) when running locally. nng.tcp does timeout and nng.ws fails with Connection refused. Both errors are pre-existing on main, indicating environmental problems rather than code problems.

## Alternatives or future work

As an alternative to this change, it was considered to add full validation of allowed properties on both reception and sending for all MQTT message types. This was not implemented at this time as it is not needed to fix the problem at hand, but it could be done instead or on top of this for more strict validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MQTT 5 Will Property validation.
  * Invalid Will Properties now return the appropriate malformed-packet error.
  * Disconnect messages now include the correct reason code for malformed packets.
  * Will Delay Interval is handled correctly after successful validation.

* **Tests**
  * Added coverage for valid and invalid MQTT Will Property combinations.

### DIFF
--- a/include/nng/protocol/mqtt/mqtt_parser.h
+++ b/include/nng/protocol/mqtt/mqtt_parser.h
@@ -111,6 +111,7 @@ NNG_DECL nng_msg *nano_encode_publish_msg(uint8_t proto_ver, uint8_t qos,
     property *prop, const char *topic, const char *topic_suffix);
 // TODO : check duplicated declaration
 NNG_DECL reason_code check_properties(property *prop, nng_msg *msg);
+NNG_DECL reason_code check_will_properties(property *prop);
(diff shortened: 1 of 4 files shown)


JaylinYu (approved):
well done

PR #1636, opened 2026-09-08
fix(tls): allow a PEM certificate with a PKCS#11 key

pbestler:
Follow-up to #1628.

`open_config_own_cert()` currently rejects any configuration in which only one of
the certificate and the key is a PKCS#11 URI:

```c
if (cert_pkcs11 != key_pkcs11) {
        log_error("NNG-TLS-CFG-OWNCHAIN"
                  "PKCS#11 strict mode: cert and key must both be PKCS#11 URIs");
        rv = NNG_EINVAL;
        goto error;
}
```

That rules out the most common HSM deployment, where the certificate stays a PEM
file in the filesystem and only the private key lives on the token.

Nothing in the engine actually requires the two to share a source:

- both branches of the function already resolve the certificate and the key
  independently;
- `open_store_uri()` performs the provider check for whichever credential is a
  URI, so the OpenSSL 3 / `pkcs11` provider requirement still holds when only the
  key is one;
- the PIN callback is installed before either credential is loaded;
- `SSL_CTX_check_private_key()` still verifies that the certificate and the key
  belong together, so a genuine mismatch is caught with `NNG_ECRYPTO`.

This drops the pairing check and covers the mixed case in `pkcs11_test`.

## Testing

Built with `-DNNG_ENABLE_TLS=ON -DNNG_TLS_ENGINE=open
-DNNG_TEST_PKCS11_PROVIDER=ON -DNNG_REQUIRE_PKCS11_PROVIDER=ON`; no new warnings
in the changed files.

`tls_test` shows the same five failures before and after this change — `tls large
message`, `tls garbled cert`, and the three PSK cases — verified by rebuilding and
rerunning against the unmodified parent commit. They are pre-existing gaps in the
OpenSSL engine and unrelated to this change. `tls mutual auth` stays green.

The new `test_pkcs11_mixed_credentials` is gated on an optional
`NNG_PKCS11_CERT_PEM` naming the PEM certificate that belongs to
`NNG_PKCS11_KEY_URI`. It returns early when that fixture is unset, so existing
token setups are unaffected.

One point worth a reviewer's attention: `SSL_CTX_check_private_key()` needs the
public part of the token key to be readable. `pkcs11-provider` exposes it in the
normal case, but a token holding a private-only object without a public
counterpart could trip this. That is exactly the path the new test exercises, and
I could not verify it against real hardware here.

## Summary by CodeRabbit

* **Bug Fixes**
  * TLS configurations can now independently use PEM data or PKCS#11 URIs for certificates and private keys.
  * Certificate and key compatibility is validated before establishing a secure connection.

* **Tests**
  * Added coverage for mixed PEM certificate and PKCS#11 key configurations.

### DIFF
--- a/src/supplemental/tls/openssl/openssl.c
+++ b/src/supplemental/tls/openssl/openssl.c
@@ -1057,7 +1057,15 @@ out:
 #endif
 }
 
-/** Configures a certificate and key from matching PEM data or PKCS#11 URIs. */
+/**
+ * Configures our identity certificate and private key.
+ *
+ * The certificate and the key are resolved independently: each may be
+ * either PEM data or a PKCS#11 URI. This allows a certificate to stay in
+ * the filesystem while its key lives on a token, which is the common HSM
+ * deployment. OpenSSL confirms that the two belong together via
+ * SSL_CTX_check_private_key() below.
+ */
 static int
 open_config_own_cert(nng_tls_engine_config *cfg, const char *cert,
     const char *key, const char *pass)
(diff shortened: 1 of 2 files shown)


pbestler:
My PR https://github.com/nanomq/NanoNNG/pull/1637  fixes this failing tests so maybe merge that first.

JaylinYu (approved):
thanks for your responsibility. OpenSSL is only an alternative ecryption layer, we didnt put too much attention on the open-sources ver.

PR #1638, opened 2026-09-09
fixes #1619 OpenSSL TLS server can complete before flushing the final…

### LINKED ISSUE #1619: OpenSSL TLS server can complete before flushing the final handshake flight

JaylinYu:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TLS handshake reliability and error reporting during network reads and writes.
  - TLS configuration now validates minimum and maximum protocol versions.
  - Invalid or oversized ALPN protocol entries are rejected consistently.
  - Invalid CA certificate configuration is reported clearly.
  - MQTT CONNECT packets containing a password without a username are rejected.
  - Invalid broker-to-client MQTT PUBLISH properties are removed before delivery, while unknown properties continue to produce a protocol error.

### DIFF
--- a/include/nng/protocol/mqtt/mqtt_parser.h
+++ b/include/nng/protocol/mqtt/mqtt_parser.h
@@ -111,7 +111,7 @@ NNG_DECL nng_msg *nano_encode_publish_msg(uint8_t proto_ver, uint8_t qos,
     property *prop, const char *topic, const char *topic_suffix);
 // TODO : check duplicated declaration
 NNG_DECL reason_code check_properties(property *prop, nng_msg *msg);
-NNG_DECL reason_code check_out_pub_properties(property *prop);
+NNG_DECL reason_code sanitize_out_pub_properties(property *prop);
 NNG_DECL property *decode_buf_properties(uint8_t *packet, uint32_t packet_len,
     uint32_t *pos, uint32_t *len, bool copy_value);
 NNG_DECL property *decode_properties(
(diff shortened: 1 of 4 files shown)


JaylinYu:
@coderabbitai review

JaylinYu:
@coderabbitai review

JaylinYu:
@coderabbitai review

wanghaEMQ (commented):
Cool. Maybe we need to add more logs to handshake()

PR #1643, opened 2026-09-11
Present the full certificate chain from certfile

D-r-P-3-p-p-3-r:
open_config_own_cert() only loaded the first certificate found in certfile (the leaf) via SSL_CTX_use_certificate(); any intermediate CA certificates concatenated after it in the same file were parsed into the BIO and then silently discarded.

The server's outgoing chain would still often look complete, but only because OpenSSL falls back to auto-completing it from whatever is in the verify cert store -- which is populated by cacertfile, a setting whose documented purpose is verifying the *peer's* certificate, not supplying the server's own chain. This made the server's presented chain depend on the (unrelated) cacertfile/verify_peer configuration, and differ between listeners depending on incidental cacertfile content, e.g. a listener with a client-cert-verification-only CA file would accidentally get a complete chain while an otherwise-identical listener without cacertfile configured would present the leaf alone and fail client-side chain validation (verify error 20: unable to get local issuer certificate).

Fix: after loading the leaf certificate, read any remaining certificates out of the same certfile buffer and add them via SSL_CTX_add_extra_chain_cert(), so the server's chain is always fully determined by certfile alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - PEM-based TLS certificates now present the complete certificate chain to clients when additional certificates are included.
  - Errors encountered while loading the certificate chain are reported appropriately.

### DIFF
--- a/src/supplemental/tls/openssl/openssl.c
+++ b/src/supplemental/tls/openssl/openssl.c
@@ -91,6 +91,7 @@ print_hex(char *str, const uint8_t *data, size_t len)
 #include <openssl/bio.h>
 #include <openssl/ssl.h>
 #include <openssl/err.h>
+#include <openssl/pem.h>
 #if !defined(LIBRESSL_VERSION_NUMBER) && \
     OPENSSL_VERSION_NUMBER >= 0x30000000L
 #define NNG_OPENSSL_HAVE_PKCS11 1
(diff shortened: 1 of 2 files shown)


JaylinYu:
@coderabbitai review

JaylinYu (approved):
I understand your case, which is the same issue we got in SDV,  But I would recommend to modify the certs file first.

D-r-P-3-p-p-3-r:
> But I would recommend to modify the certs file 
> first.

Can you elaborate what you mean with that?
We migrated from Mosquitto to nanomq. And the cert stack file that worked before failed.

If I used the option `cacertfile` for the named listener it worked - by chance (see desciption above). But without it the broker only sent one of the three certs.

PR #1644, opened 2026-09-11
Add per-listener mount_point support

D-r-P-3-p-p-3-r:
Add a per-listener mount_point string that flows: HOCON config ->
conf_tcp/conf_tls/conf_websocket struct field -> a new NANO_MOUNT_POINT nng_listener option (mirrors the existing NANO_CONF option) -> copied into each transport's endpoint (ep) struct -> copied into each new pipe (p->mount_point) at pipe-match time -> stamped onto the pipe's conn_param (cparam->mount_point, an owned copy) right after the CONNECT packet is parsed via the new conn_param_apply_mount_point().

This lets multiple tenants/apps share one broker's topic tree without collision: SUBSCRIBE/UNSUBSCRIBE filters and PUBLISH topics get the publishing/subscribing client's own mount_point prepended on ingress; PUBLISH delivery (and retained-message delivery on SUBSCRIBE) strip the recipient's own mount_point on egress. $SYS topics/filters always bypass both directions, and $share/<group>/ shared-subscription filters get the mount_point inserted after the group name.

Startup validation (conf_mount_point_validate()) rejects empty mount_points, leading $, or +/# wildcards.

This is the NNG-side (transport/config/conn_param plumbing) of the feature; the matching nanomq/ (broker application) call sites and topic-rewrite logic are on the identically-named branch drp3pp3r/listener-mount-point in the nanomq repo.

### DIFF
--- a/include/nng/nng.h
+++ b/include/nng/nng.h
@@ -1659,6 +1659,9 @@ NNG_DECL void  nng_msg_set_conn_param(nng_msg *msg, void *ptr);
 NNG_DECL const char    *conn_param_get_clientid(conn_param *cparam);
 NNG_DECL const uint8_t *conn_param_get_pro_name(conn_param *cparam);
 NNG_DECL const void *   conn_param_get_will_topic(conn_param *cparam);
+NNG_DECL const char    *conn_param_get_mount_point(conn_param *cparam);
+NNG_DECL void           conn_param_set_mount_point(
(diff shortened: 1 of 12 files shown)


JaylinYu:
@coderabbitai review

JaylinYu:
@coderabbitai review

PR #1657, opened 2026-09-18
prevent duplicated retain msg delivery

JaylinYu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate retained messages with the same publish topic from appearing in collected results.
  * Ensured vector values are evaluated only once during insertion, improving reliability when expressions have side effects.

* **API Updates**
  * Updated retained-message lookup to provide results through an output vector parameter. Applications using this public interface may need corresponding integration changes.


JaylinYu:
SQLite retain may needs more attention

xinyi-xs (commented):
Thumbs up!

